PR #331, opened 2014-09-17
compact prefix -/+ instead of verbose ago/ahead sulfix.

gcb:
so i can have a few more chars for the log message on a small screen. ...i think i can probably work this out as an option if nobody likes this way (i.e. everyone works on 42" monitors :)


jonas:
Added something similar to what what proposed in this PR.

PR #380, opened 2015-02-23
Improve C99 compliance for wider compiler support

hashpling:
This range of commits improves C99 compliance and as a result allows tig to be compiled with the native compilers on Solaris (SunStudio cc) and AIX (xlc).

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -6,6 +6,8 @@ master
 
 Improvements:
 
+ - Improve C99 compliance so Tig compiles with the native compilers on
+   Solaris (SunStudio cc) and AIX (xlc). (GH #380)
  - Add move-half-page-up and move-half-page-down actions. (GH #323)
  - Preserve the cursor position when changing the diff context. 
  - Add option to turn off automatic enabling of `--topo-order` when the graph is
--- a/src/io.c
+++ b/src/io.c
@@ -498,12 +498,13 @@ io_get(struct io *io, struct buffer *buf, int c, bool can_read)
 bool
 io_write(struct io *io, const void *buf, size_t bufsize)
 {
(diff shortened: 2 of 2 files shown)


jonas:
Wow, thanks. Nice patch set. The travis build shows two warnings which are trivial to fix. Unless you want to update the pull request, I can do these two touch ups as well as adding an entry to the NEWS file before merging.

```
src/prompt.c:860:21: warning: suggest braces around initialization of subobject
      [-Wmissing-braces]
                struct key key = {0};
                                  ^
                                  {}

src/main.c:96:26: warning: suggest braces around initialization of subobject
      [-Wmissing-braces]
        struct commit commit = {0};
                                ^
                                {}
```

hashpling:
I guess it's looking for `struct commit commit = {{0}};` or even `{""};`. To me, plain `{0}` is idiomatic C and "double bracing" looks a bit ugly. Note that this warning should go away in a newer version of gcc: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119 .

hashpling:
So I tested with `{{0}}` in both places and both native compilers are happy so it's up to you. If a warning-free travis is important to you then I don't object to these touch ups even though I think it's a bogus warning in this case.

jonas:
What is important is that `make all-debug` which adds `-Wall` to `CFLAGS` works for me locally and it does not. So I will add the missing braces.

jonas:
Thanks merged.

It's probably time to embrace more C99 constructs in the code base, especially for struct initialization. This is a good start!

hashpling:
Thanks! I'll test with `all-debug`, at least on Linux, for any further patches which I come up with.

PR #385, opened 2015-02-28
Avoid initializing ncurses before parsing options

johnkeeping:
If called before a view has been set up, report() will call die() which
prints to stderr and exits.  However, because init_display() is called
well before the first view is set, this is likely to result in endwin()
being called which will clear the user's terminal meaning that they do
not see the message.

For example, running "tig blame master" results in:

	user@host ~/src/tig $ tig blame master
	user@host ~/src/tig $d options to blame

	tig 2.0.3 (Dec 20 2014)

	Usage: tig        [options] [revs] [--] [paths]
	   or: tig log    [options] [revs] [--] [paths]
	   or: tig show   [options] [revs] [--] [paths]
	   or: tig blame  [options] [rev] [--] path
	   or: tig grep   [options] [pattern]
	   or: tig refs
	   or: tig stash
	   or: tig status
	   or: tig <      [git command output]

	Options:
	  +<number>       Select line <number> in the first view
	  -v, --version   Show version and exit
	  -h, --help      Show help message and exit

with the cursor positioned on the second prompt above most of the error
output.

### DIFF
--- a/src/tig.c
+++ b/src/tig.c
@@ -634,8 +634,6 @@ main(int argc, const char *argv[])
 	if (load_refs(FALSE) == ERR)
 		die("Failed to load refs.");
 
-	init_display();
-
 	if (pager_mode)
 		request = open_pager_mode(request);
 
--- a/src/view.c
+++ b/src/view.c
@@ -755,9 +755,15 @@ split_view(struct view *prev, struct view *view)
 void
 maximize_view(struct view *view, bool redraw)
 {
+	static bool display_initialized;
+
 	memset(display, 0, sizeof(display));
 	current_view = 0;
 	display[current_view] = view;
+	if (!display_initialized) {
+		init_display();
+		display_initialized = true;
+	}
 	resize_display();
 	if (redraw) {
 		redraw_display(FALSE);


jonas:
What terminal does this happen with?

In any case, I think it would be best to put the lazy initialization code in `reset_display()`.

johnkeeping:
I see the problem with xterm.

Do you mean `resize_display()`?  I can't find `reset_display()`...  I expect you're right that it would be better elsewhere, do you want me to update the PR or are you happy to apply a change of your own?

jonas:
Yes, sorry, I meant `resize_display()`. I will amend the fix when I merge it. Thanks.

jonas:
It looks like the problem is that `done_display()` and therefore `endwin()` is called twice when `die()` is called. First via `die_callback` (which always points to `done_display()` once `init_display()` has been called), and again by the `atexit()` handler also installed once `init_display()` has been called.

`atexit()` must be there so that ncurses is "shutdown" when Tig exits due to a signal. I therefore propose the following patch which fixes the issue for me.
```
diff --git a/src/display.c b/src/display.c
index 82d903c..f77bc78 100644
--- a/src/display.c
+++ b/src/display.c
@@ -411,7 +411,9 @@ report(const char *msg, ...)
 static void
 done_display(void)
 {
-	endwin();
+	if (cursed)
+		endwin();
+	cursed = FALSE;
 }
 
 void
```
Note that in order to reproduce this issue I had to test in a fresh terminal; once Tig has run it looks like there is some code that makes the bug go away.

jonas:
When you have time could you please give this a try?

johnkeeping:
That fixes it for me.  Thanks!

jonas:
Thanks for testing @johnkeeping

PR #391, opened 2015-03-16
Fix insertion/ordering of refs in refs_by_id map.

### LINKED ISSUE #390: no display of remote refs if tag exists
tig shows remote refs only if there is no tag for this commit. Bisect shows that commit 26c858d introduces this problem.

To reproduce use the small shell script below. It outputs for me:

    OK   2015-03-16 16:04 Heiko Schlichting I [master] {origin/master} {origin/HEAD} commited
    FAIL 2015-03-16 16:04 Heiko Schlichting I [master] <Marke> commited

```
#!/bin/sh

check() {
    grep -q origin/master $1 && echo -n "OK   " || echo -n "FAIL "
    head -1 $1
}

TDIR=$(mktemp -d)
mkdir $TDIR/fern
cd $TDIR/fern
git init > /dev/null
date > file
git add file > /dev/null
git commit -m "commited" > /dev/null
cd $TDIR
git clone fern nah > /dev/null 2>&1
cd $TDIR/nah
TIG_SCRIPT=$TDIR/tig.script
export TIG_SCRIPT
OUT=$TDIR/tig.display
cat > $TIG_SCRIPT << EOF
:save-display $OUT
:quit
EOF
tig
check $OUT
git tag Marke
tig
check $OUT
echo "For cleaning up remove $TDIR"
```

AlexanderS:
During insertion of a ref into the refs_by_id map the code lost some refs because the next pointer in the linked list get modified in the wrong way. This seems to fix #390.

### DIFF
--- a/src/refdb.c
+++ b/src/refdb.c
@@ -137,19 +137,19 @@ add_ref_to_id_map(struct ref *ref)
 		ref->next = NULL;
 	}
 
-	ref->next = *ref_lists_slot;
-	*ref_lists_slot = ref;
+	if (*ref_lists_slot == NULL || ref_compare(ref, *ref_lists_slot) <= 0) {
+		ref->next = *ref_lists_slot;
+		*ref_lists_slot = ref;
+	}
+	else {
+		struct ref *refs;
+		for (refs = *ref_lists_slot; refs->next; refs = refs->next) {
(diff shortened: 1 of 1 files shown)


jonas:
Thanks. I added Heiko's test case to the commit.

Issue #404, opened 2015-04-24
populate %(file) when cursor over diffstat information in commit view


dustymabe:
When my cursor is over a filename in the diffstat (like below) then can we have %(file) be populated for calling external tig bindings? If I have my cursor over the actual diff of the file then %(file) is populated, but not when I have it over the diffstat.

afile.txt | 16 ++++++++--------

tig version: 2.0.3

PR #405, opened 2015-04-26
Assign variables from diffstat

vivien:
Hi Jonas!

This pull request adds a new `diff_find_header_from_stat()` function used in `diff_common_enter()` and `diff_common_select()` in order to assign the %(file) variable from a diff stat line, which contains the (possibly truncated) filename.

Also use these variables from `diff_common_edit()` so we can now open the editor from a diff stat line.

Fixes the not found issue #404.


jonas:
Awesome. Thanks @vivien !

PR #409, opened 2015-05-23
Gh 5 incremental search

jasonchang2082:


jonas:
Hi @jasonchang2082 , one of the reasons I didn't merge this was due to minor bugs that had to be sorted out first. Are you planning to fix those? Also, this changes all searching to be incremental. Not sure but we might want to make that optional. What do you think?

edi9999:
Yes probably make it optional, but make it set by default :-)

jonas:
@jasonchang2082 I'm closing this, since I don't see much point in merging a PR you have created from one of my branches if you don't plan to get involved.

PR #429, opened 2015-08-18
check if mem is NULL in DEFINE_ALLOCATOR

ideal:
Or if mem is NULL, it will cause segmentation fault.


jonas:
Good catch. Looks like there's a couple of flaws in the logic of that macro. Will merge and likely see if I can clean it up later.

PR #433, opened 2015-08-27
Add zsh completion file for autoload

wkentaro:


wkentaro:
Moved from https://github.com/Homebrew/homebrew/pull/43320

aemonge:
+1

jonas:
@wkentaro Thanks. I suggest that the file is renamed to `contrib/tig-completion.zsh`. I can do this if you prefer.

wkentaro:
In that case, the completion file should be renamed to `_tig` at installation.
Where can I find the code which installs the completion files? (`make install-doc` seems not)

jonas:
The completion files are not currently installed. I suggest putting that information in the file header and then publishing a patch to Homebrew adding a line after the [bash completion install](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/tig.rb#L34) using [zsh_completion.install](https://github.com/Homebrew/homebrew/blob/8ba134eda537d2cee7daa7ebdd9f728389d9c53e/Library/Formula/gibo.rb#L10).

wkentaro:
Okay, updated.

jonas:
Thanks

PR #434, opened 2015-08-28
Fix vertical split artifact

swegener:
39acb5ab03a6e3b02eaeca6c9387301615e1524e introduced a regression, resulting in the vertical split separator still being displayed when switching to horizonal split.

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -167,6 +167,27 @@ redraw_display_separator(bool clear)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


vivien:
I didn't noticed this bug! Confirmed and tested with `vertical-split = auto`.

Tested-by: Vivien Didelot <vivien.didelot@gmail.com>

jonas:
@swegener Thanks for the patch.
@vivien Thanks for confirming.

PR #438, opened 2015-09-10
Fix testframework

mjg:
At least in my shell(s), "cd foo" echoes the new cwd to stdout so that
"cd foo && pwd" echoes the new cwd twice, which lets the test framework
fail completely (due to the wrong $source_dir and everything that
depends on it).

Make that construct "cd foo >/dev/null && pwd" to be super safe, rather
than simplifying to "cd foo" and relying on the behaviour of "cd".

(I'm actually wondering how that ever worked ;)

### DIFF
--- a/test/tools/libtest.sh
+++ b/test/tools/libtest.sh
@@ -21,7 +21,7 @@ if [ -n "${BASH_VERSION:-}" ]; then
 fi
 
 test="$(basename "$0")"
-source_dir="$(cd "$(dirname "$0")" && pwd)"
+source_dir="$(cd $(dirname "$0") >/dev/null && pwd)"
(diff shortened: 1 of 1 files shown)


mjg:
All tests pass over here (fedora 21), of course. Dunno what Travis is doing ;)

jonas:
Thanks, I'll rebase once I've fixed the built in `master`.

jonas:
Merged as b46a06d959da1f78a7467a91b46aa727071f169c. I took the freedom to clarify a bit in the commit title.

Thanks again.

mjg:
Thanks, I should know how to write better commit subjects, sorry ;)

PR #439, opened 2015-09-15
blame: allow user to specify rev arguments to blame

peff:
The options parser for blame is currently quite strict, and can handle only a single positive revision in opt_rev_args. But there are several parsing problems with this:

 1. If you run `tig blame --reverse foo.c`, the `--reverse`     flag ends up in opt_rev_args (filter_options knows it's a     flag, but since it is a revision flag, it gets put     there).

    We try to access `--reverse.c:foo.c`, which is     nonsensical. Moreover, if you did specify a start-point     like `HEAD foo.c`, tig will complain that you have more     than one revision arg.

 2. If you run `tig blame --since=123`, this is converted by     `rev-parse --revs-only` to `--max-age=123`, and is     included in opt_rev_args, and you run into the same     problems as above.

 3. If you run `tig blame HEAD~20..HEAD`, the commit range     is split into two arguments, `HEAD` and `^HEAD~20`, both     of which end up in opt_rev_args. Similarly, the user can     specify this `split` form themselves.

We can fix this by further parsing `opt_rev_args`. Instead of assuming that it must contain zero or one arguments, we can pick out the "positive" ref as our starting point (and complain if there is more than one).

The rest of the rev-args we add to `%(blameargs)`, so that git-blame can act on them.

### DIFF
--- a/src/blame.c
+++ b/src/blame.c
@@ -93,17 +93,35 @@ blame_open(struct view *view, enum open_flags flags)
 
 	if (is_initial_view(view)) {
 		/* Finish validating and setting up blame options */
-		if (!opt_file_args || opt_file_args[1] || (opt_rev_args && opt_rev_args[1]))
+		if (!opt_file_args || opt_file_args[1])
 			usage("Invalid number of options to blame");
-
-		if (opt_rev_args) {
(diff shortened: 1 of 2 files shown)


jonas:
Awesome, thanks.

PR #457, opened 2015-11-22
add a %(text) variable

vivien:
This exposes the text of the currently selected line as the %(text)
variable. Only the pager view is supported at the moment.

This is particularly handy for using Tig as a scriptable pager. For
instance, here's a dumb remote manager:

    $ echo 'bind pager <Enter> git remote show -n %(text)' >> ~/.tigrc
    $ echo 'bind pager U git remote update -p %(text)' >> ~/.tigrc
    $ git remote | tig

Pressing Enter shows remote information, pressing U fetches the remote.


jonas:
Cool.

PR #461, opened 2015-12-07
Update Vim bindings to include <C-f> and <C-b>

alexandercampbell:
These bindings are commonly used in vim and other text viewers (for example, `less`) to scroll pages at a time.

Thanks!

### DIFF
--- a/contrib/vim.tigrc
+++ b/contrib/vim.tigrc
@@ -22,6 +22,9 @@ bind generic gn view-next
 bind main    G  none
 bind generic G  move-last-line
 
+bind generic <C-f> move-page-down
+bind generic <C-b> move-page-up
+
 bind generic v  none
 bind generic vm view-main
 bind generic vd view-diff


alexandercampbell:
@jonas any ideas about fixing the test failure?

jonas:
I'll try to find some time to look into this over the weekend.

The vim keybindings gets tested by test/help/user-command-test and the
problem is that the formatted text is wider than the screen used by the
test resulting in truncation when the ncurses screen is dumped to file. The
fix is likely to just increate the width by setting COLUMNS higher.

On Thu, Dec 10, 2015 at 12:12 PM, Alexander Campbell <
notifications@github.com> wrote:

> @jonas <https://github.com/jonas> any ideas about fixing the test failure?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jonas/tig/pull/461#issuecomment-163690518>.
>



-- 
Jonas Fonseca

jonas:
Thanks. Updated the tests and merged in 2401021b93d64445fefa3c3002a030ad209fd3a2

PR #468, opened 2016-01-08
Update INSTALL.adoc

samiraguiar:
Add instructions to install on Windows.


netei:
Hi, out of curiosity, how do you "install the packages `git`, `gcc-core`, `make`, `libiconv-devel`
+and `libncurses-devel`" ? Is they anything such as `apt-get` ?

samiraguiar:
@netei not really, no. You run setup.exe again, which gives you a list of packages to install/update/remove.

PR #478, opened 2016-02-19
Set view->parent to NULL when view is closed.

egnagflow:
Fixes an issue where diff full screen diff view up/down navigation results
in moving the underlying commit rather than the diff line if the diff view
had been opened in split format (Enter key) before.


jonas:
Thanks.

Issue #472, opened 2016-01-17
Empty unstaged changes shown on rename


proski:
Run this file in an empty directory to reproduce the bug:
```
#! /bin/sh

rm -rf .git
rm -f file*
git init
echo data >file1
git add file1
git commit -m "commit1"
echo data >file2
git add file2
git rm file1
```
Run tig after that script. It will show "Unstaged changes" in the change list, but pressing Enter on that entry won't have any effect. It is important for reproducing the bug that `file1` and `file2` have the same contents.

PR #491, opened 2016-04-06
Prevent staged rename from displaying unstaged changes

harish:
#472 - Took a whack at this. Appears that git status --porcelain -z outputs "R dst\0src\0" for a rename.

### DIFF
--- a/src/repo.c
+++ b/src/repo.c
@@ -137,6 +137,7 @@ index_diff(struct index_diff *diff, bool untracked, bool count_all)
 	};
 	struct io io;
 	struct buffer buf;
+	unsigned skip_count = 0;
 	bool ok = true;
 
 	memset(diff, 0, sizeof(*diff));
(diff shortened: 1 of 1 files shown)


src/repo.c
```diff
@@ -137,14 +137,25 @@ index_diff(struct index_diff *diff, bool untracked, bool count_all)
...
-	while (io_get(&io, &buf, 0, true) && (ok = buf.size > 3)) {
+	while (io_get(&io, &buf, 0, true) && ((ok = buf.size > 3) || skip_count)) {
+
+		if (skip_count > 0){
+			skip_count--;
+			continue;
+		}
+
+		/* Skip source filename in rename */
+		if (buf.data[0] == 'R')
```
jonas: Shouldn't renames increase the `diff->staged` count? This patch will report the index as clean.

jonas:
I think it would make sense to add a test for this

src/repo.c
harish: This will still get incremented at line 152 as it did before. The next iteration where it reads in the 2nd filename will skip. Originally I used a counter because I thought it'd hit two filenames. Would it be more clear to use a flag like 'skip_next' or 'last_was_rename'?

harish:
Sure, I'll give it a shot. I'm having a little trouble running tests on OS X-- everything seems to fail. Any tips?

make test
./config.status config.h
config.status: creating config.h
config.status: config.h is unchanged
      TEST  test/blame/default-test
            | Failed 4 out of 0 test(s)
            | [FAIL] blame-default.screen not found
            | [FAIL] blame-with-diff.screen not found
            | [FAIL] blame-with-diff-no-file-filter.screen not found
            | [FAIL] blame-parent-of-74537d9.screen not found

jonas:
@harish Thanks for the patch. I updated it and added some test cases. Regarding the problem with `make test` do you have GNU sed installed?

src/repo.c
jonas: Yes, I missed that. I ended up putting an explicit `io_get` call to read the source filename since `ok = buf.size > 3` was breaking for short filenames.

harish: I believe the || skip_count should've caught that case but I definitely like your version better-- much cleaner. Thanks :)

jonas: Yes, it would have continued but `ok` would have become `false` due to the tricky assignment in the while condition causing the result of index_diff to be discarded

PR #492, opened 2016-04-07
Show signature in diff view

MichaelAquilina:
With the recent introduction of GPG signing support in Github, I feel like this is a good time to introduce support for showing signatures in tig.

Currently I've added this in the diff and log view. 

I couldn't figure out where options are defined to make this an option the use can enable and disable - was hoping for some feedback on that.


MichaelAquilina:
Closing this PR. You can add the following to tigrc instead:
```
set log-options = --show-signature
set diff-options = --show-signature
```

PR #494, opened 2016-04-12
Make 'make configure' quieter

mopsfelder:
With this patch, 'make configure' is quieter:

$ make configure
       GEN  configure

And if V=1 is specified, then the output is more verbose:

$ make V=1 configure
./autogen.sh
./autogen.sh: running: aclocal -I tools
./autogen.sh: running: autoconf --include=tools
./autogen.sh: running: autoheader --include=tools


jonas:
Thanks

Issue #460, opened 2015-11-30
tig status: Add "branch up-to-date with remote"


ffes:
When I use `git status` it shows if your current branch is up-to-date with the remote.

```
On branch master
Your branch is up-to-date with 'origin/master'.
```
or
```
On branch master
Your branch is ahead of 'origin/master' by 1 commit
```

`tig status` does not show this information. Simply adding it to the first line would be great:
```
On branch master. Your branch is up-to-date with 'origin/master'.
```

netei:
I personnaly like this idea

ffes:
For some reason this one was not automatically closed when the #504 was merged.

PR #504, opened 2016-06-05
Show tracking info in tig status

rgburke:
- Updated `tig status` to show tracking info when appropriate as suggested in #460.
- The `git rev-list` command is used to determine the difference with the remote branch instead of parsing the output of `git status`, as there's no guarantee the format of the latter will remain the same in future.
- Added new test `on-branch-tracking-info-test` based on `on-branch-test` to help ensure this new functionality works as expected.

Below is a screenshot showing the appearance of this feature:

tig-branch-tracking-info

### DIFF
--- a/src/status.c
+++ b/src/status.c
@@ -199,6 +199,71 @@ status_restore(struct view *view)
 	clear_position(&view->prev_pos);
 }
 
+static bool
+status_branch_tracking_info(char *buf, size_t buf_len, const char *head,
+			    const char *remote)
+{
+	if (!string_nformat(buf, buf_len, NULL, "%s...%s",
+			    head, remote)) {
+		return false;
+	}
+
+	const char *tracking_info_argv[] = {
+		"git", "rev-list", "--left-right", buf, NULL
+	};
+
+	struct io io;
+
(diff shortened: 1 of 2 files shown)


ffes:
As the one who suggested #460, I did some simple testing and what I tested all works as expected. Thanks for implementing it! I really hope this PR gets merged.

Closes #460

jonas:
I like this. The status line might get a bit long but at least it is clearly formulated.

src/status.c
```diff
@@ -199,6 +199,71 @@ status_restore(struct view *view)
...
+			    const char *remote)
+{
+	if (!string_nformat(buf, buf_len, NULL, "%s...%s",
+			    head, remote)) {
+		return false;
+	}
+
+	const char *tracking_info_argv[] = {
+		"git", "rev-list", "--left-right", buf, NULL
+	};
```
peff: Sorry I'm a little late to the party, but I happened to notice this while pulling master. You can actually do `git rev-list --left-right --count` to get just the two counts. Probably not a big deal, but it would save a little pipe and parsing overhead.

PR #505, opened 2016-06-07
Fix drawing long lines in draw_text_expanded

### LINKED ISSUE #500: diff view corruption while scrolling right
## Steps to Reproduce:

1. Run the following commands:

    git clone -b bug-replicate https://github.com/h3xx/tig.git bug-temp ; cd bug-temp ; TIGRC_USER=/dev/null tig

2. Open the commit at the HEAD in the diff panel, either in split view or full diff view.

3. Use the right arrow key to scroll right about 1000 characters.

The screen corruption looks like this. The green text shouldn't be there because the only line long enough to be shown there is the red deletion line.

screenshot-2016-0506-1655

Tig version ebd8a238266cdebb970ec8d77a3382963d437e09

rgburke:
- Adjust max_width on each iteration to take account of the column width that has been drawn.
 - This issue only manifested itself on very long lines which required more than one invocation of draw_chars i.e. when `length > sizeof(text)`.
 - This fixes #500.

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -113,11 +113,13 @@ draw_text_expanded(struct view *view, enum line_type type, const char *string, i
 
 	do {
 		size_t pos = string_expand(text, sizeof(text), string, length, opt_tab_size);
+		size_t col = view->col;
 
 		if (draw_chars(view, type, text, -1, max_width, use_tilde))
 			return true;
 		string += pos;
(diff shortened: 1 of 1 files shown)


jonas:
Awesome thanks

h3xx:
This fixes the issue I reported. Thanks!

PR #506, opened 2016-06-10
Added support for move to parent in main view.

### LINKED ISSUE #388: Feature request - support "move to parent" action for the main view
Take below graph for example, under "Main view", I thought that when I press `,` (move to parent) key on the view of `b0d721f` it will navigate to `c1d1f71`, but not happen :( 
If "parent" key can support on "Main view", that will be truly awesome and handy.

Though this is not a good example, since this 2 commit are really near.
However in some case, it's really far away.

image

And also to handle a *Merge* commit as below

```
commit cafafe9e7dcd69157bb50d185ca13dc3f42d2873 (HEAD, origin/master, origin/HEAD, master)
Merge: bcc4484 b0d721f
Author: Jonas Fonseca <jonas.fonseca@gmail.com>
Date:   Wed Mar 11 20:00:48 2015 -0400

    Merge pull request #386 from andreasstieger/master

    Update FSF addesss in COPYING
```
Should provide somewhat like below I guess for a more convenient navigation
1: jump-to-first-parent
2: jump-to-second-parent

rdnetto:
Fixed #388.

### DIFF
--- a/src/tig.c
+++ b/src/tig.c
@@ -266,7 +266,11 @@ view_driver(struct view *view, enum request request)
 		break;
 
 	case REQ_PARENT:
-		report("Moving to parent is not supported by the %s view", view->name);
+		if (! strcmp(view->name, "main")) {
+			move_view(view, request);
+		} else {
+			report("Moving to parent is not supported by the %s view", view->name);
+		}
 		break;
 
 	case REQ_BACK:
--- a/src/tig.c
+++ b/src/tig.c
@@ -266,11 +266,7 @@ view_driver(struct view *view, enum request request)
 		break;
 
 	case REQ_PARENT:
-		if (! strcmp(view->name, "main")) {
-			move_view(view, request);
-		} else {
-			report("Moving to parent is not supported by the %s view", view->name);
-		}
+		report("Moving to parent is not supported by the %s view", view->name);
 		break;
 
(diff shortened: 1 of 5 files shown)


src/tig.c
```diff
@@ -266,7 +266,11 @@ view_driver(struct view *view, enum request request)
 		break;
 
 	case REQ_PARENT:
-		report("Moving to parent is not supported by the %s view", view->name);
+		if (! strcmp(view->name, "main")) {
+			move_view(view, request);
```
jonas: The changes to support `REQ_PARENT` should go in `src/main.c` which is all the main view specific code. There's a `main_request` method where the `case REQ_PARENT` and related `move_view` changes in this patch can exist.

rdnetto:
I've moved the code into `main_request()`, and modified it to reuse an existing method instead of duplicating the scrolling logic.

I also added a special case for treating HEAD as the parent of Unstaged Changes.

jonas:
Thanks, I ended up massaging the code a bit myself and also added a test and a NEWS entry.

chenkaie:
That would be nicer if support 2 direction movement, a really intuitive musle memory when I doing the movement :)
- `,`              : move to parent
- `SHIFT + ,` : move to child

PR #508, opened 2016-06-21
add support for worktress

nmorey:
Worktree pass is complex as it used a combination of gitlinks, commondir and config.
The simplest way is to simply use the git rev-parse --show-toplevel which returns the top of the worktree

Signed-off-by: Nicolas Morey-Chaisemartin nmorey@kalray.eu

---

This is not a pretty patch but it solves the worktree issue and also work in standard trees and bare repo.
Unless there is a use case where core.worktree is set but git rev-parse --show-toplevel cannot find the worktree (or finds a diffrent one), some thing like this should handle evrey case


jonas:
It looks like this breaks at least one test. I'll take a look eventually, but probably not until I get around to release version 2.2.

PR #509, opened 2016-07-01
Add command line aliases for `tig refs`

ffes:
I find myself typing `tig branch` to show the branches a lot. Therefore I added a command line option as an alias for `tig refs`. I also added `tig tag` for completeness.

I assume there is a (minor) change needed for bash completion but I could find out how to do that myself. Hopefully someone with more knowledge of bash completion can fix that.

### DIFF
--- a/doc/tig.1.adoc
+++ b/doc/tig.1.adoc
@@ -56,6 +56,12 @@ log::
 refs::
 	Start up in refs view.
 
+branch::
+	Start up in refs view. Here for compatibility with the 'git branch' command.
+
+tag::
+	Start up in refs view. Here for compatibility with the 'git tag' command.
+
 stash::
 	Start up in stash view.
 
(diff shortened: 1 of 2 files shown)


jonas:
I am still thinking whether this is better solved by adding something similar to git aliases. In any case the patch should also update the help message before being merged.

ffes:
I indeed forgot about the help message. And it is missing from the synopsis of the man page as well.

What do you like the usage to look like?

```c
"   or: tig refs|branch|tag\n"
```

or 

```c
"   or: tig refs\n"
"   or: tig branch\n"
"   or: tig tag\n"
```

or 

```c
"   or: tig refs\n"
"       tig branch\n"
"       tig tag\n"
```

I have a slight preference for the first.

jonas:
I'll go ahead and close this. The idea is good but is missing the ability to filter by the ref type.

PR #512, opened 2016-07-21
Fix error when running with --silent.

Ferada:
Otherwise running `autogen.sh` manually, or using `make configure
--silent` will result in errors:

    ./autogen.sh: 16: test: =: unexpected operator


jonas:
Thank you

Issue #513, opened 2016-07-25
Respect the XDG standard for configuration files


rpdelaney:
Rather than reading the user configuration from `~/.tigrc` by default, instead check `$XDG_CONFIG_HOME/tig/config` and only then look for `~/.tigrc`. Further reading: https://ploum.net/207-modify-your-application-to-use-xdg-folders/

PR #514, opened 2016-08-06
options: Respect the defaults specified in XDG Basedir

grawity:
<https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html>:

    "If $XDG_CONFIG_HOME is either not set or empty, a default equal to
    $HOME/.config should be used."

Ref: #513

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -978,18 +978,21 @@ load_options(void)
 			return error("Error in built-in config");
 	}
 
-	if (!tigrc_user) {
+	if (tigrc_user) {
+		load_option_file(tigrc_user);
+	} else {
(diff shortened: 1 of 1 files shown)


jonas:
Thanks for fixing this oversight. Unless you get around to it I'd like to add a test for this behaviour as well.

I also realize that I forgot to update the manpage with information about this new environment variable.

grawity:
I'm not sure if the manuals should speak about `$XDG_CONFIG_HOME/tig/config` or `~/.config/tig/config`? The former is more accurate (and also what git-config(1) uses), but it's also more confusing to new readers...

jonas:
Thanks, merged

PR #517, opened 2016-08-16
Bind Ctrl-D and Ctrl-U to half-page movements by default

JelteF:
Scrolling half pages shows the direction of movement much better than full pages, because not the entire screen is different.
This is also the default in vim and in less.


jonas:
Thanks for this contribution. Applied.

PR #525, opened 2016-09-18
Add navigation between merge commits

phi-gamma:
Allow skipping forward and backward between merge commits by means of
two actions “move-next-merge” and “move-prev-merge”. The search
considers all lines containing the merge symbol.

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -210,7 +210,7 @@ find_next_merge(struct view *view, enum request request)
 		return error("Unknown search request");
 	}
 
-	if (!view->matched_lines && find_merges(view))
+	if (!view->matched_lines && !find_merges(view))
 		return ERROR_OUT_OF_MEMORY;
 
 	code = find_next_match_line(view, direction, false);
--- a/src/search.c
+++ b/src/search.c
@@ -207,7 +207,7 @@ find_next_merge(struct view *view, enum request request)
 		break;
 
 	default:
-		return error("Unknown search request");
+		return error("Invalid request searching for next merge");
 	}
 
 	if (!view->matched_lines && !find_merges(view))
--- a/src/search.c
+++ b/src/search.c
@@ -69,7 +69,7 @@ find_merges(struct view *view)
 		if (!view->ops->get_column_data(view, line, &column_data))
 			continue;
 
-		if (!column_data.graph->is_merge(canvas))
+		if (column_data.graph && !column_data.graph->is_merge(canvas))
 			continue;
 
 		if (!realloc_unsigned_ints(&view->matched_line, view->matched_lines, 1))
--- a/src/search.c
+++ b/src/search.c
@@ -15,6 +15,8 @@
 #include "tig/prompt.h"
 #include "tig/display.h"
 #include "tig/draw.h"
+#include "tig/main.h"
+#include "tig/graph.h"
 
 DEFINE_ALLOCATOR(realloc_unsigned_ints, unsigned int, 32)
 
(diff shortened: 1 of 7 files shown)


phi-gamma:
I’ve been wanting this feature since the first time I use tig ;) I consider
it immensely useful when browsing larger projects like the kernel:
Pull Linus’ tree first thing in the morning an flip between merges to
get a broad idea of what changed since your last fetch.

src/search.c
```diff
@@ -159,6 +191,44 @@ find_next(struct view *view, enum request request)
...
+
+	case REQ_MOVE_PREV_MERGE:
+		direction = -1;
+		break;
+
+	default:
+		return error("Unknown search request");
+	}
+
+	if (!view->matched_lines && find_merges(view))
```
jonas: This will always return `ERROR_OUT_OF_MEMORY` when `find_merges` succeeds.

src/search.c
```diff
@@ -159,6 +191,44 @@ find_next(struct view *view, enum request request)
...
+	case REQ_MOVE_NEXT_MERGE:
+		direction = 1;
+		break;
+
+	case REQ_MOVE_PREV_MERGE:
+		direction = -1;
+		break;
+
+	default:
+		return error("Unknown search request");
```
jonas: I think this error message should be changed.

src/tig.c
```diff
@@ -292,6 +292,11 @@ view_driver(struct view *view, enum request request)
 		find_next(view, request);
 		break;
 
+	case REQ_MOVE_NEXT_MERGE:
+	case REQ_MOVE_PREV_MERGE:
+		find_merge(view, request);
```
jonas: This will segfault on views that does not have a graph. I suggest this only reports an error:

```c
report("Moving between merge commits is not supported by the %s view", view->name);
```

And the `find_merge` cases are moved to `main_request` in `src/main.c`.

src/search.c
```diff
@@ -53,6 +55,36 @@ find_matches(struct view *view)
...
+	for (lineno = 0; lineno < view->lines; lineno++) {
+		struct line *line = &view->line[lineno];
+		struct commit *commit = line->data;
+		struct graph_canvas *canvas = &commit->graph;
+		struct view_column_data column_data;
+
+		if (!view->ops->get_column_data(view, line, &column_data))
+			continue;
+
+		if (!column_data.graph->is_merge(canvas))
```
jonas: `column_data.graph` will be NULL here for all views except `main`.

src/search.c
```diff
@@ -53,6 +55,36 @@ find_matches(struct view *view)
...
+			continue;
+
+		if (!realloc_unsigned_ints(&view->matched_line, view->matched_lines, 1))
+			return false;
+
+		view->line[lineno].search_result = true;
+		view->matched_line[view->matched_lines++] = lineno;
```
jonas: Any particular reason to reuse these member that are usually for search highlighting?

phi-gamma:
···<date: 2016-10-17, Monday>···<from: Jonas Fonseca>···

> Thanks for this contribution. I think it is a nice feature,
> however, there're a couple of things I think are needed before
> this can be merged. Let me know if you want me to polish things
> up.

I very much appreciate the review. I’ll try to have something
ready till the end of the week. So far I’ve addressed most of
your points but I’m not sure how to proceed with this one:

> > +		struct line *line = &view->line[lineno];
> +		struct commit *commit = line->data;
> +		struct graph_canvas *canvas = &commit->graph;
> +		struct view_column_data column_data;
> +
> +		if (!view->ops->get_column_data(view, line, &column_data))
> +			continue;
> +
> +		if (!column_data.graph->is_merge(canvas))
> +			continue;
> +
> +		if (!realloc_unsigned_ints(&view->matched_line, view->matched_lines, 1))
> +			return false;
> +
> +		view->line[lineno].search_result = true;
> +		view->matched_line[view->matched_lines++] = lineno;
> 
> Any particular reason to reuse these member that are usually for search highlighting?

No particular reason except to have the code as close to the
search routines as possible.

The two obvious options would be a) to extend ``struct view``
with similar fields for collecting the merges, and b) to find the
next merge on demand. Personally I’d go with the latter because
it’s less complex. Any thoughts?

Best,
Philipp

phi-gamma:
I’ve addressed each of your remarks in a single commit. After review
give me a note and I’ll squash them into a single changeset.

jonas:
Thanks squashed and merged.

PR #533, opened 2016-10-14
Add A as a binding to apply a stash without dropping it.

parkercoates:
Pretty straight forward.

### DIFF
--- a/tigrc
+++ b/tigrc
@@ -153,6 +153,7 @@ set mouse-scroll		= 3		# Number of lines to scroll via the mouse
 
 bind main	C	?git cherry-pick %(commit)
 bind status	C	!git commit
+bind stash	A	?git stash apply %(stash)
 bind stash	P	?git stash pop %(stash)
 bind stash	!	?git stash drop %(stash)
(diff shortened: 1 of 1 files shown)


parkercoates:
Okay, this commit changes the contents of the help screen output, since a new binding has been added. This breaks the tests on the help output. Is the expectation that I manually update the samples used in the help tests, or is there a fancy process for such changes?

jonas:
@parkercoates If you rebase on top of current `master` you should be able to run this command to update the expected test output

```bash
make test/help/all-keybindings-test && \
  cp test/tmp/help/all-keybindings-test/help-default.screen test/help/all-keybindings-test.expected
```

jonas:
OK, ended up merging and fixing the test myself. Thanks for the contribution and suggestion to easily allow the expected output to be updated.

parkercoates:
That's the real trick to open source. ;) Offer a one line pull request, then sit back and watch as it prompts the reviewer into writing an additional 130 lines of supporting changes!

Thank you very much!

Issue #313, opened 2014-07-21
Add support for diff-highlight


dwijnand:
I would be really nice if tig supported [diff-highlight](https://github.com/git/git/tree/v2.3.5/contrib/diff-highlight).

dwijnand:
So.. I was wondering- how viable is including this? Is it possible to "shell out to perl" for this?

I've negligible experience in C, so I could use any help I can get in implementing this.

Thanks

jonas:
There's two things required to make this work. First, the diff view will
need to learn how to parse ASCII control characters. To start with, only
the ones to enter/exit highlighting are required. Since the diff view is
line-based, `struct line` will need a flag to indicate whether `diff_draw`
needs to "interpret" the line characters.

Next, is the actual running of `diff-highlight` as a sort of filter of the
underlying `git-diff`/`git-show` command. Tig will need to set up the pipe
using the IO module. And also have hooks to kill/close the filter process.
It is not supported right now.

danielmanesku:
It would be really nice to have this feature (or even better [diff-so-fancy](https://github.com/so-fancy/diff-so-fancy#screenshot)), it saves a lot of time.
It also got some light from git team: https://github.com/blog/2188-git-2-9-has-been-released

tqmz:
Would be nice indeed.

wikimatze:
+1 for that feature

jonas:
<img width="532" alt="screen shot 2016-11-07 at 00 29 31" src="https://cloud.githubusercontent.com/assets/8417/20047361/620ccf7c-a481-11e6-8e43-ffe00a76afe5.png">

dwijnand:
What am I doing wrong that I see the following?

![image](https://cloud.githubusercontent.com/assets/344610/20051025/9f0edb20-a4c4-11e6-92ad-6ab51dcc4d0a.png)

jonas:
How did you configure it?
```
09:17:19 /opt/tig#master > head $(which diff-highlight)
#!/usr/bin/perl

use 5.008;
use warnings FATAL => 'all';
use strict;

# Highlight by reversing foreground and background. You could do
# other things like bold or underline if you prefer.
my @OLD_HIGHLIGHT = (
	color_config('color.diff-highlight.oldnormal'),
09:17:33 /opt/tig#master > grep diff-highlight ~/.tigrc
set diff-highlight = true
```
[...]

PR #537, opened 2016-11-09
Fix broken diff-highlight on some platforms

dtwlin:
Fix issue with diff-highlight that the env variable is not null terminated

Refs #313

Signed-off-by: David Lin <dtwlin@google.com>


jonas:
Thanks!

Issue #540, opened 2016-11-15
update bash completion (like git's current)


lbonn:
The bash completion script have been adapted from its counterpart in git.

However, it seems that it now lags a bit behind in terms of functionalities.

In particular, I often use tig to show the changes introduced by a branch: `tig ..branch_example` and being able to type `tig ..br<tab>` would be handful (this works with `git log` in current versions)

This may also apply to zsh for some other patterns but this particular use case seems to be covered.

jonas:
Yes, I agree that it could need an overhaul.
Help with this would be really appreciated!

lbonn:
I made a lousy attempt in the train last week but I didn't have a doc on bash completion's internals with me.

I'll try again if I have some free time in the next few weeks.

rohieb:
This should now be solved by #960.

PR #541, opened 2016-11-20
completion: Simple rewrite attempt

lbonn:
Work towards #540.

I made a small completion script which re-uses the base git's. It is supposed to have better support for revisions and options completions.

The drawback is that the base git's completion must be installed beforehand. For most bash users, this should be a given. For zsh, it would probably won't work in this state but tig's completion is already builtin in zsh (see [git's completion in zsh's tree](https://sourceforge.net/p/zsh/code/ci/master/tree/Completion/Unix/Command/_git)). It only matches `git log` behaviour though...

It should also be checked for compatibility with old git versions. Does tig have a minimum git version requirement?

A standalone version should not be too hard to write with heavy copy-pasting if needed.

Either way, this first draft can be used to validate its correct behaviour. I tried some common combinations but additional testing would be nice.

### DIFF
--- a/contrib/tig-completion.bash
+++ b/contrib/tig-completion.bash
@@ -27,7 +27,7 @@
 #       which may be slightly slower.
 #
 
-__tig_cmds="blame status show log stash grep"
+__tig_cmds="blame status show log refs stash grep"
 
 __tig_complete_revlist_or_cmd ()
 {
--- a/contrib/tig-completion.bash
+++ b/contrib/tig-completion.bash
@@ -15,262 +15,125 @@
 #
 # To use these routines:
 #
-#    1) Copy this file to somewhere (e.g. ~/.tig-completion.sh).
-#    2) Added the following line to your .bashrc:
+#    1) Install and activate the base git completion
+#    2) Copy this file to somewhere (e.g. ~/.tig-completion.sh).
+#    3) Added the following line to your .bashrc:
 #        source ~/.tig-completion.sh
 #
-#    3) You may want to make sure the git executable is available
+#    4) You may want to make sure the git executable is available
 #       in your PATH before this script is sourced, as some caching
 #       is performed while the script loads.  If git isn't found
 #       at source time then all lookups will be done on demand,
 #       which may be slightly slower.
 #
 
-__tigdir ()
-{
-	if [ -z "$1" ]; then
-		if [ -n "$__git_dir" ]; then
-			echo "$__git_dir"
-		elif [ -d .git ]; then
-			echo .git
-		else
-			git rev-parse --git-dir 2>/dev/null
-		fi
-	elif [ -d "$1/.git" ]; then
-		echo "$1/.git"
-	else
-		echo "$1"
-	fi
-}
+__tig_cmds="blame status show log stash grep"
 
-_tigcomp ()
+__tig_complete_revlist_or_cmd ()
 {
-	local all c s=$'\n' IFS=' '$'\t'$'\n'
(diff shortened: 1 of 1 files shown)


jonas:
Cool, I was not able to get the `_completion_loader` working on mac os x. Will take a closer look when I get more time.

I have one suggestion to improve blame completion to also match revisions:

```
diff --git i/contrib/tig-completion.bash w/contrib/tig-completion.bash
index a40088c..deba0cb 100755
--- i/contrib/tig-completion.bash
+++ w/contrib/tig-completion.bash
@@ -29,9 +29,10 @@
 
 __tig_cmds="blame status show log stash grep"
 
-__tig_complete_revlist_or_cmd ()
+__tig_complete_revlist_or_list ()
 {
 	local pfx ls ref cur_="$cur"
+	local list="$1"
 	case "$cur_" in
 	*..?*:*)
 		return
@@ -83,7 +84,7 @@ __tig_complete_revlist_or_cmd ()
 		__gitcomp_nl "$(__git_refs)" "$pfx" "$cur_"
 		;;
 	*)
-		__gitcomp "$__tig_cmds"
+		__gitcomp "$list"
 		__gitcomp_nl_append "$(__git_refs)"
 		;;
 	esac
@@ -118,7 +119,7 @@ _tig ()
 			__gitcomp "-v -h"
 			;;
 		*)
-			__tig_complete_revlist_or_cmd ;;
+			__tig_complete_revlist_or_list "$__tig_cmds" ;;
 		esac
 		return
 	fi
@@ -131,7 +132,7 @@ _tig ()
 	status)
 		_git_status;;
 	blame)
-		__gitcomp "$(__git_complete_file)";;
+		__tig_complete_revlist_or_list "$(__git_complete_file)";;
 	*)
 		_git_log ;;
 	esac
```

lbonn:
About OSX: maybe I overestimated the portability of this approach.

I also tried your patch but if I just type `tig blame <tab><tab>`, I only get a list of refs which is not what I need from blame in most of the cases. It only gives a list of files if the current word does not match any rev (eg: if I have a file `mouse.c` and a branch `master`, only `master` would appear if I type `m<tab>`).

`__git_complete_file` might be the cause, here is its definition along with `__git_complete_revlist` in git-completion.bash.

```
__git_complete_file ()
{
	__git_complete_revlist_file
}

__git_complete_revlist ()
{
	__git_complete_revlist_file
}
```

Apparently this is used to match a revision in priority and only files if none matches, typically for `log`, `diff` for which it is recommended to use `--` to limit to files.

Is it common to supply a revision to tig blame? Even if the completion to files and revs works, having the full list of revs (which can be pretty large in some repos) if the user only seeks files is not really helpful.

But maybe I just don't know about the use cases of `tig blame rev`.

Also, the whole `blame` completion seems to be missing from git's script to let it fall back on bash's default completion to file. But this is probably an argument from authority :).

lbonn:
I put all the mainline completion utilities we depend on directly in the file to avoid being too dependent on the environment and/or git's completion version (just like it was before).

In theory, the inlining could be semi-automatized (with some effort...) to ease the maintenance but we probably don't need to update the completion too often.

The issue you raised about the blame completion still need to be resolved. Personally, I'd still be in favor of using the mainline approach which favors the most common command usage.

jonas:
Thanks, I will install it and test it out next week

lbonn:
News on this?

FWIW, I have been using it almost daily since then.

Noteworthy examples:

    tig ..<tab>  # shows refs on this branch, files on stock completion
    tig master -- <tab>  # shows files on this branch, refs on stock completion

PR #546, opened 2016-11-26
Update ax_with_curses.m4 to support pkg-config

jlec:
Latest version of ax_with_curses.m4 supports pkg-config. This is important
as libcurses is split into libcurses.so and libtinfo.so on many distros
and linking against libcurses.so is not sufficiant. Pkg-config allows
proper detection of necessary libs.

Signed-off-by: Justin Lecher <jlec@gentoo.org>


configure.ac
```diff
@@ -21,15 +21,23 @@ AC_CHECK_FUNCS([mkstemps], [AC_SUBST([NO_MKSTEMPS], ["#"])])
...
-AX_WITH_CURSES
-case "$ax_cv_ncurses" in "no")
-	AC_MSG_ERROR([ncurses not found])
-esac
-case "$ax_cv_ncursesw" in "no")
-	AC_MSG_WARN([The found ncurses library does not support wide-char.])
-	AC_MSG_WARN([This means that tig will not correctly render UTF-8.])
-esac
-AC_SUBST(CURSES_LIB)
+PKG_CHECK_MODULES([CURSES], [ncursesw],
```
jonas: I ended up reverting this change because the `AX_WITH_CURSES` macro will call `pkg-config` automatically.

jonas:
Great, thanks @jlec. Hope this still works on Gentoo with the changes I made

jlec:
Works perfectly fine.

PR #547, opened 2016-11-28
Fix author ident cache being keyed by email only

minaguib:
This addresses the issue where an author name change "wins" and tig
shows it for all historical commits by same email under old names

Addresses jonas/tig/issues/526 and jonas/tig/issues/424


jonas:
Cool, thanks Mina. Will take a closer look tonight.

minaguib:
@jonas Unfortunately the test suite (even on vanilla `master` branch) is entirely broken on my mac, so I didn't realize my changes in my branch are causing failures until travis turned it red above.

I'll try to figure out why `make test` on master is failing on my mac and, if I can get it to a green state, will do another push that fixes test failures failing directly due to my branch's changes.

jonas:
It's not well documented but the test suite requires GNU sed to run _and_ it must be symlilinked to `sed`.

jonas:
Thanks again for this contribution

PR #579, opened 2017-03-22
Add smart-case option

### LINKED ISSUE #320: "Smart" case-sensitivity during search
It would be very nice if we could specify an option so that searching is always done case-insensitive.

Even better would be an auto option that is case-insensitive if the search expression is all lowercase, and case-sensitive otherwise (this is how my emacs works)

hluk:
New option enables case-insensitive search only if current search
expression doesn't contain any uppercase letters.

This is simple solution which checks only uppercase ASCII letters.

Fixes #320

### DIFF
--- a/doc/tigrc.5.adoc
+++ b/doc/tigrc.5.adoc
@@ -281,6 +281,11 @@ The following variables can be set:
 
 	Ignore case in searches. By default, the search is case sensitive.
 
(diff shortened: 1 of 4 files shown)


jonas:
Thanks, merging as is, but I might try to unify the `ignore_case` and `smart_case` options to allow

    set ignore_case = smart

Issue #581, opened 2017-03-28
Keybinding for navigating by line number?


benjmin-r:
For longer commits (and probably also any other view which can become quite long like the log view), it would be very helpful to be able to navigate more precisely, based on line numbers. Similar to how a `3j` in vim moves 3 lines down, and `4k` moves 4 lines up. 
A different way to approach this, would be to be able to specify a line number, which one can jump to, similar to how `30gg` in vim jumps to line 30 of the current buffer.

I can't find already a keybinding for this? I wasn't able to find one. 

Thanks a lot for any help/pointer and most of all for tig. It's super useful to me! ✨

jonas:
Number prefixes are like you say not supported. You can jump to a specific line with `:123`.

benjmin-r:
Thanks for the clarification. The `:123` syntax is already quite useful. 

Would it make sense to add this capability to the manual. Possibliy under the section `Cursor Navigation` or as an addition to the explanation of `:` under `Misc` that's already there? I could make a PR for that, if you deem it sensible.

jonas:
Yes, please, that would be much appreciated

PR #584, opened 2017-03-29
Explain command for jumping to specific line

benjmin-r:
As discussed in #581 this adds a short note to the manual, that mentions `:` can also be used to jump to line.


jonas:
Thank you.

PR #597, opened 2017-05-27
Improve blame argument behavior

m42a:
This moves the checking for `-C` to after `opt_blame_options` and (as per the tigrc man page) only sets `opt_blame_options` via the command line if command line arguments were actually given.


jonas:
Thanks!

PR #600, opened 2017-06-03
Replace status-untracked-dirs command in tigrc

wendorf:
Commented sample binding for status-untracked-dirs is updated to the
non-deprecated status-show-untracked-dirs option.


jonas:
Thanks.

PR #605, opened 2017-06-19
fix graphs where the same commit is hit by a merge and a branch

zevv:
Trying to fix graphs where the same commit is hit by a merge and a new branch, but the graph gives the impression the commit is not merged.

The old case rendered this result. Not that the bottom branch looks
like it terminates without being merged:

````
 ● Update NEWS with post 2.1.1 changes
 ●─╮ Merge pull request #457 from vivien/text-variable
 │ ● add a %(text) variable
 ●─╯ Merge pull request #439 from peff/fix-blame-args
 │ ● blame: allow user to specify rev arguments to blame
 ●─╯ Update OSX make config to find brew installed ncurses
 ● Remove unneeded calls to {def,reset}_prog_mode
````

With fix I get this:

````
 ● Update NEWS with post 2.1.1 changes
 ●─╮ Merge pull request #457 from vivien/text-variable
 │ ● add a %(text) variable
 ●─┤ Merge pull request #439 from peff/fix-blame-args
 │ ● blame: allow user to specify rev arguments to blame
 ●─╯ Update OSX make config to find brew installed ncurses
 ● Remove unneeded calls to {def,reset}_prog_mode
````

Related to https://github.com/jonas/tig/issues/419

My current solution is not nice: it depends on the order of evaluation in the
graph_symbol_to_XXX() functions so that the new case of graph_symbol_merge_branch()
is hit before graph_symbol_turn_left() or graph_symbol_merge() do.

The graph-generating code is complex, and I have not spent enough time in it to
know for sure my fix does not mess up other situations.

### DIFF
--- a/src/graph-v2.c
+++ b/src/graph-v2.c
@@ -859,6 +859,15 @@ graph_symbol_multi_merge(const struct graph_symbol *symbol)
 	return true;
 }
 
+
+static const bool
+graph_symbol_merge_branch(const struct graph_symbol *symbol)
+{
+
+	return graph_symbol_turn_left(symbol) && graph_symbol_merge(symbol);
+}
+
+
 static const bool
 graph_symbol_vertical_bar(const struct graph_symbol *symbol)
 {
@@ -949,6 +958,9 @@ graph_symbol_to_utf8(const struct graph_symbol *symbol)
 	if (graph_symbol_vertical_bar(symbol))
 		return " │";
 
+	if (graph_symbol_merge_branch(symbol))
+		return "─┤";
+
 	if (graph_symbol_turn_left(symbol))
 		return "─╯";
 
@@ -1000,6 +1012,10 @@ graph_symbol_to_chtype(const struct graph_symbol *symbol)
 	} else if (graph_symbol_vertical_bar(symbol)) {
 		graphics[0] = ' ';
 		graphics[1] = ACS_VLINE;
+
+	} else if (graph_symbol_merge_branch(symbol)) {
+		graphics[0] = ACS_HLINE;
+		graphics[1] = ACS_RTEE;
 
 	} else if (graph_symbol_turn_left(symbol)) {
 		graphics[0] = ACS_HLINE;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by zevv** [comment]
Oh look there's a `make test` target. Back to the drawing board for me. Still, any feedback appreciated!

**Comment by jonas** [comment]
I must admit it has been a while since I looked at the graph code. As you can see it is the second attempt of rendering the commit graph and the work which was quite an improvement over v1 was done by Benjamin Bergman in 87009cb789d861b302e46bc451e31b44a4d932ab.

From what I remember it keeps a state of three rows (prev, current, next). Each time a commit is inserted together with a list of its parents, the "next" row is updated, then the current row is drawn and state is updated by moving "current" to "prev". The rendering itself is a bunch of heuristics which needs to take into account the limitations of drawing the graph with commits on each line which makes the routing quite complex. Not sure if it helps, but that is what comes to mind.

**Comment by jonas** [comment]
The easiest way to test is either to create a test file based on what is inside `test/graph/` and then run the test by passing the path to make, e.g. `make test/graph/00-simple-test`.

The graph tests themselves can be defined mimicking the output of `git-log`:

```
commit A B
    A with parent B
commit B C D
    B is a merge of C and D   
commit C E
    C has E as parent
commit D E
    D also has E as parent
commit E
    Initial commit
```

zevv:
Quoting Jonas Fonseca (2017-06-20 03:05:39)
 
> From what I remember it keeps a state of three rows (prev, current,
> next). Each time a commit is inserted together with a list of its
> parents, the "next" row is updated, then the current row is drawn and
> state is updated by moving "current" to "prev". The rendering itself
> is a bunch of heuristics which needs to take into account the
> limitations of drawing the graph with commits on each line which
> makes the routing quite complex. Not sure if it helps, but that is
> what comes to mind.

Yeah, that helps. The 'bunch of heuristics' is where I go wrong, I need
to grasp the semantics of a lot of those 'symbols' attached to each
commit before I can make the proper conclusion about the right graph
characters.

I'll see how far I can get, thanks for the explanation.



--
:wq
^X^Cy^K^X^C^C^C^C

jonas:
Superseded by #605 curtesy of @swegener

PR #606, opened 2017-06-19
also interpret button5 as wheel-down

### LINKED ISSUE #321: Scrolling up works, but not down
Currently on an unstable Debian, I'm using tig (version 2.0.2) in tmux (version 1.9) in xterm (version 308).

In my `.gitconfig` I have:

```
[tig]
    mouse = true
```

Moving with the arrow keys and JK works as excepted. Scrolling with the mouse wheel to go up also works as expected. However scrolling down with the mouse wheel does nothing...

Mind you, my mouse wheel works well everywhere else, be it in vim or tmux to scroll buffers (up AND down).

So, do you have any idea of what can cause this?

Thanks :)

rolandwalker:
fixes #321 .  Likely this should always have been button5, instead of button2, per https://github.com/mirror/ncurses/blob/af3d0ee323cbb22d2a7596d564bf68f7307f5076/ncurses/base/lib_mouse.c#L1025

### DIFF
--- a/src/tig.c
+++ b/src/tig.c
@@ -589,7 +589,11 @@ handle_mouse_event(void)
 	if (!view)
 		return REQ_NONE;
 
+#ifdef BUTTON5_PRESSED
+	if (event.bstate & (BUTTON2_PRESSED | BUTTON5_PRESSED))
+#else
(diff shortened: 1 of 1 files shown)


jonas:
Thanks, makes sense to me.

PR #607, opened 2017-06-19
add action view-close-no-quit

rolandwalker:
unbound by default


jonas:
Outstanding, very clean and +100 for both updating docs and adding a test.

PR #609, opened 2017-06-21
improve test suite portability

rolandwalker:
Tested under dash/shellcheck/os x; hopefully will also pass Travis.

This PR gets `make test` within 1 failure of passing under OS X.  (There is a dependency on GNU sed within `test/stage/split-chunk-test`.)

### DIFF
--- a/test/tools/libtest.sh
+++ b/test/tools/libtest.sh
@@ -283,10 +283,10 @@ show_test_results()
 		sed "s/^/$indent[skipped] /" < .test-skipped
 		return
 	fi
-	if [ -n "$trace" -a -n "$TIG_TRACE" -a -e "$TIG_TRACE" ]; then
+	if [ -n "$trace" ] && [ -n "$TIG_TRACE" ] && [ -e "$TIG_TRACE" ]; then
 		sed "s/^/$indent[trace] /" < "$TIG_TRACE"
 	fi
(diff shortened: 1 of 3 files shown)


test/tools/libtest.sh
```diff
@@ -356,6 +356,10 @@ test_require()
 		diff-highlight)
 			diff_highlight_path="$(git --exec-path)/../../share/git-core/contrib/diff-highlight/diff-highlight"
 			if [ ! -e "$diff_highlight_path" ]; then
+				# alt path
+				diff_highlight_path="$(git --exec-path)/../../share/git/contrib/diff-highlight/diff-highlight"
```
jonas: 👍

test/tools/libtest.sh
```diff
@@ -299,7 +299,7 @@ show_test_results()
 		failed="$(grep FAIL < .test-result | wc -l)"
 		count="$(sed -n '/\(FAIL\|OK\)/p' < .test-result | wc -l)"
 
-		printf "Failed %d out of %d test(s)%s\n" $failed $count
+		printf "Failed %d out of %d test(s)\n" "$failed" "$count"
```
jonas: Great catch.

jonas:
Thanks, manually squashed. Also fixed `show_results.sh` to work with POSIX shell.

PR #610, opened 2017-06-21
add test coverage for `source` command

rolandwalker:


jonas:
Great to have more coverage for `source`.

PR #611, opened 2017-06-21
WIP remove "expected" test outputs during setup

rolandwalker:
I needed this change, or something similar, for a pending PR with multiple testcases in one script — and it seems sensible.

I'm not 100% certain that `$prefix` doesn't also touch `expected/stdin`.

PR #612, opened 2017-06-22
add a -q parameter to source for source-if-present

rolandwalker:
won't pass test suite without ~#611~ ~#615~ #615 was incorporated here.


jonas:
Thanks for incorporating #615

PR #613, opened 2017-06-22
Make build reproducible

andreasstieger:
Make build reproducible
https://reproducible-builds.org/


jonas:
Thanks. I was not aware of this effort.

PR #614, opened 2017-06-22
remove dependency on GNU sed newline handling in test suite

rolandwalker:
The issue is that POSIX sed always fills in a trailing newline, whereas GNU sed maintains the input more faithfully.

This is a small file and it seems just as easy to write it with a literal `printf`.

After #609 this will be sufficient for the test suite to run on OS X and likely any BSD derivative.

### DIFF
--- a/test/stage/split-chunk-test
+++ b/test/stage/split-chunk-test
@@ -17,7 +17,7 @@ steps '
 '
 
 in_work_dir create_dirty_workdir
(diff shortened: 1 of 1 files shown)


rolandwalker:
Note: this will fail tests without quoting of `$@` from #609 .

jonas:
Thanks I squashed this into the #609 PR

PR #615, opened 2017-06-22
Add per-testcase params `--tigrc` and `--assert-stderr`

rolandwalker:
Added for #612, replacing #611.

It is more convenient to capture/test `source` warnings deriving from a `.tigrc` than from a script.

### DIFF
--- a/test/tools/libtest.sh
+++ b/test/tools/libtest.sh
@@ -492,7 +492,7 @@ test_case()
 	echo "$name" >> test-cases
 	cat > "$name.expected"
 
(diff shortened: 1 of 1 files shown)


rolandwalker:
You'd rather these formed as a single commit?

jonas:
Yes if you could add these to the `source -q` PR that would be great.

rolandwalker:
Done.  `git log` favors big commits; `git bisect` favors small commits.

jonas:
Agree. I usually prefer small commits, but lost track of the changes in the test portability PR by looking at them commit wise.

PR #616, opened 2017-06-24
doubledash and similar test suite linting

rolandwalker:
Not sure you will agree with all of these.  A few cases border on POSIX matters but most are defensive driving WRT whitespace and punctuation.  No functional change.

* doubledash before constructs which may lead with dash
* `cat <` idiom similar to doubledash
* further quoting to avoid word splitting
* prefer `find ... -exec` over `xargs`. `xargs` splits on space and `xargs -0` is non-POSIX
* `while read -r` over `for ... $(cat ...)` to split on linefeed not space
* `mkdir` with `-p` unless `mkdir` is subject of test
* `$*` over `$@` when meaning is concatenation (somewhat pedantic)
* `printf` over `echo` as it is better behaved/defined
* `grep -c` over `wc -l` avoids leading whitespace
* `grep` over `sed`
* constrain grep FAIL/OK patterns to beginning-of-line
* typo in valgrind parameter key
* whitespace

### DIFF
--- a/test/tools/libtest.sh
+++ b/test/tools/libtest.sh
@@ -20,10 +20,10 @@ if [ -n "${BASH_VERSION:-}" ]; then
 	IFS=$'\n\t'
 fi
 
-test="$(basename "$0")"
-source_dir="$(cd $(dirname "$0") >/dev/null && pwd)"
-base_dir="$(echo "$source_dir" | sed -n 's#\(.*/test\)\([/].*\)*#\1#p')"
-prefix_dir="$(echo "$source_dir" | sed -n 's#\(.*/test/\)\([/].*\)*#\2#p')"
+test="$(basename -- "$0")"
+source_dir="$(cd "$(dirname -- "$0")" >/dev/null && pwd)"
+base_dir="$(printf '%s\n' "$source_dir" | sed -n 's#\(.*/test\)\([/].*\)*#\1#p')"
+prefix_dir="$(printf '%s\n' "$source_dir" | sed -n 's#\(.*/test/\)\([/].*\)*#\2#p')"
 output_dir="$base_dir/tmp/$prefix_dir/$test"
 tmp_dir="$base_dir/tmp"
 output_dir="$tmp_dir/$prefix_dir/$test"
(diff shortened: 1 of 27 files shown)


jonas:
Still thinking about this one.

test/tools/libtest.sh
```diff
@@ -399,20 +399,20 @@ test_setup()
...
 valgrind_exec()
 {
-	kernel="$(uname -s 2>/dev/null || echo unknown)"
+	kernel="$(uname -s 2>/dev/null || printf 'unknown\n')"
 	kernel_supp="test/tools/valgrind-$kernel.supp"
 
 	valgrind_ops=
 	if [ -e "$kernel_supp" ]; then
-		valgrind_ops="$valgrind_ops --suppresions='$kernel_supp'"
+		valgrind_ops="$valgrind_ops --suppressions='$kernel_supp'"
```
jonas: Oh, was the parameter renamed?

test/tools/libtest.sh
```diff
@@ -310,16 +310,16 @@ show_test_results()
 		# Replace CR used by Git progress messages
 		tr '\r' '\n' < .test-result
 	elif [ "$verbose" ]; then
-		count="$(sed -n '/\(OK\)/p' < .test-result | wc -l)"
-		printf "Passed %d assertions\n" "$count"
+		count="$(grep -c '^ *\[OK\]' < .test-result || true)"
```
jonas: 👍

test/tools/show-results.sh
```diff
@@ -21,10 +21,10 @@ if [ -n "${BASH_VERSION:-}" ]; then
...
-tests="$(find test/ -name ".test-result" | wc -l)"
-asserts="$(find test/ -name ".test-result" | xargs grep '\[\(OK\|FAIL\)\]' | wc -l)"
-failures="$(find test/ -name ".test-result" | xargs grep FAIL | wc -l || true)"
-skipped="$(find test/ -name ".test-skipped" | wc -l || true)"
+tests="$(find test/ -name ".test-result" | grep -c . || true)"
+asserts="$(find test/ -name ".test-result" -exec cat -- "{}" \; | grep -c '^ *\[\(OK\|FAIL\)\]' || true)"
+failures="$(find test/ -name ".test-result" -exec cat -- "{}" \; | grep -c '^ *\[FAIL\]' || true)"
+skipped="$(find test/ -name ".test-skipped" | grep -c . || true)"
 
 if [ "$((failures))" = 0 ]; then
```
jonas: Maybe my workaround to use `$((...))` can be removed?

test/tools/libtest.sh
rolandwalker: It may accept both - I pasted that spelling from the man page.

jonas:
Thanks for rebasing, will merge and likely cause a conflict for #664.

PR #617, opened 2017-06-24
cleanups and tests on view format widths

rolandwalker:
* `width` on `line-number` is documented, accepted, but ignored; make it work
* `width` on `status` is documented, accepted, but ignored; remove it
* `width` on `text` is undocumented, accepted, but ignored; remove it
* `width` on `commit-title` is undocumented, accepted, but ignored; remove it
* fix: `width=1` fails when column is trimmed
* add tests for all remaining columns that support `width`
* adjust `main/view-split-test` to match this logic

I simply removed status/text/commit-title because the intent of the design seems to be that these be rightmost and right-filling.  Though it is certainly a reasonable feature to want.

Edit: `view-split-test` was adjusted to fit this, though neither behavior is correct. The UTF-8 logic should be capable of truncating to two visible cells `作者`.

### DIFF
--- a/doc/tigrc.5.adoc
+++ b/doc/tigrc.5.adoc
@@ -484,8 +484,6 @@ ref::
 status::
 	- 'display' (mixed) [no|short|long|<bool>]: How to display the status
 	  label.
-	- 'width' (int): Width of the column. When set to zero, the width is
-	  automatically sized to fit the content.
 
 text::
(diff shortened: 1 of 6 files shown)


jonas:
At one point I had a prototype to make "fields" right adjustable to mimic the layout of `gitk`. As you rightfully point out that is not supported and makes the `width` column setting useless in some cases.

Thanks for the fixes.

PR #618, opened 2017-06-26
recognize empty readline input and clear report area

rolandwalker:
Reproducible (but not amenable to the test suite) by

```
:<Enter>
```

which leaves the `:` behind


jonas:
Good catch.

PR #619, opened 2017-06-26
bindable control characters Ctrl-C, Ctrl-V, Ctrl-S, Ctrl-@, hardcode readline Ctrl-C

### LINKED ISSUE #583: Use <C-C> for keybindings
Is there a way to bind `<C-C>` to a custom command instead of terminating tig?
My aim is to bind it to sth. like escape, in order to leave the prompt (see: https://github.com/jonas/tig/issues/582).

rolandwalker:
* <kbd>Ctrl-C</kbd>, <kbd>Ctrl-V</kbd>, <kbd>Ctrl-S</kbd> would previously have been bindable in the main UI only if the user did some unusual things with `stty` before executing `tig`
* switching to `raw()` mode requires a special case to preserve prior <kbd>Ctrl-Z</kbd> behavior
* <kbd>C-@</kbd> (NUL) is a valid key as far as ncurses is concerned, and useful as a translation of <kbd>Ctrl-&lt;Space&gt;</kbd>
* bindings at the readline prompt are affected by this PR in that *eg* <kbd>Ctrl-S</kbd> becomes available, but the user still must configure those keys separately in `~/.inputrc`: doc that
* <kbd>Ctrl-C</kbd> at the readline prompt is a special case and cannot be bound  in `~/.inputrc`; however we can absorb the `SIGINT` to get intuitive "abort input" behavior. 
* bind <kbd>Ctrl-C</kbd> to `quit` by default in the main UI, to emulate previous behavior

The <kbd>Ctrl-C</kbd> in readline behavior has a more natural look after #618.
Fixes #583

### DIFF
--- a/doc/tigrc.5.adoc
+++ b/doc/tigrc.5.adoc
@@ -545,6 +545,9 @@ the keybindings for the *generic* keymap, and last the default keybindings.
 Thus, the view keybindings override the generic keybindings which override the
 built-in keybindings.
 
+Keybindings at the line-entry prompt are typically governed by the readline
+library, and are configured separately in `~/.inputrc`.  See 'readline(1)'.
+
 --
 
 Keymaps::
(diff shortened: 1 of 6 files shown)


doc/tigrc.5.adoc
```diff
@@ -545,6 +545,9 @@ the keybindings for the *generic* keymap, and last the default keybindings.
 Thus, the view keybindings override the generic keybindings which override the
 built-in keybindings.
 
+Keybindings at the line-entry prompt are typically governed by the readline
+library, and are configured separately in `~/.inputrc`.  See 'readline(1)'.
```
jonas: 👍

jonas:
Thank you for the detailed commit message. My personal setup and usage of the terminal is too simple to ever have been aware of these limitations. It makes sense to use `raw` and the impact in added complexity is limited.

PR #620, opened 2017-06-27
Persistent readline history 

rolandwalker:
Persist readline history to `~/.tig_history` or `$XDG_DATA_HOME/tig/history`.

Arguably the `$XDG_DATA_HOME` part doesn't belong in `prompt.c` but this does keep `#ifdef HAVE_READLINE` tucked into a single file.

It might be preferable to persist only the "search" part of the history as `less` does.  Certainly that's the part of the feature I intend to use. However, all readline inputs are currently merged in the volatile history, and I wanted to avoid changing existing UI.

Another desirable refinement would be to take `HISTORY_SIZE` from a `~/.tigrc` option, and support disabling persistent history when set to zero.

It doesn't seem practical to provide test coverage as there isn't a reliable way to inspect the binary for its compile-time readline configuration.

### DIFF
--- a/src/prompt.c
+++ b/src/prompt.c
@@ -24,6 +24,7 @@
 #ifdef HAVE_READLINE
 #include <readline/readline.h>
 #include <readline/history.h>
+#define HISTORY_SIZE 500
 #endif /* HAVE_READLINE */
 
 static char *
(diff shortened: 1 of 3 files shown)


jonas:
Yeees, looking forward to using this. \o/
Will take a look tomorrow.

include/tig/prompt.h
```diff
@@ -46,6 +46,8 @@ struct input {
 enum input_status prompt_default_handler(struct input *input, struct key *key);
 char *read_prompt_incremental(const char *prompt, bool edit_mode, bool allow_empty, input_handler handler, void *data);
 char *read_prompt(const char *prompt);
+const char *prompt_histfile(void);
+void prompt_teardown(void);
```
jonas: One minor nit, these don't have to be exported.

src/prompt.c
```diff
@@ -466,10 +467,44 @@ read_prompt(const char *prompt)
 	return line;
 }
 
+const char *
```
jonas: ... and both can be declared `static`.

jonas:
Very nice.

src/prompt.c
```diff
@@ -478,6 +513,17 @@ read_prompt(const char *prompt)
...
+prompt_histfile()
+{
+	return "";
+}
+
+static void
+prompt_teardown(void)
+{
+}
+
```
jonas: Forgot to mention that these can be removed. But I can do that after merge.

PR #622, opened 2017-06-28
performance: throttle titlebar screen updates during loading

rolandwalker:
Instead of updating the UI at maximum granularity:

* blame view progress: update titlebar stats every 5 percent
* line-oriented views: update titlebar stats every 100 lines

This little optimization turns out to give a fair gain.  In a very large repo on SSD, the main view completes 0.4 sec faster for me (8% speedup).  Across many small repos I find the gains are always measurable, though almost always less than 0.1 sec.  These values will vary by platform, repo and particularly by terminal emulator.

Even though we are still calling `wprintw()` repeatedly, ncurses is able to notice when the screen contents do not change, and optimize away unneeded writes.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -665,7 +665,10 @@ update_view_title(struct view *view)
 		wprintw(window, " - %s %d of %zd",
 					   view->ops->type,
 					   line->lineno,
-					   view->lines - view->custom_lines);
+					   MAX(line->lineno,
+					       view->pipe
+					       ? 100 * (size_t) ((view->lines - view->custom_lines) / 100)
+					       : view->lines - view->custom_lines));
 	}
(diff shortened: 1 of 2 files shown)


src/view.c
```diff
@@ -665,7 +665,10 @@ update_view_title(struct view *view)
 		wprintw(window, " - %s %d of %zd",
 					   view->ops->type,
 					   line->lineno,
-					   view->lines - view->custom_lines);
+					   MAX(line->lineno,
```
jonas: 👍

jonas:
Heh, reminds me of the tweet about `npm set progress=false` considerably speeding up `npm install`. :)

PR #623, opened 2017-06-28
Redraw prompt after leaving it with C-C

0ax1:
Previously, '^C' was permanently shown after leaving the prompt with
C-C. This PR adds `redraw_display` to the sigint handler, in order to 
clear the prompt view after leaving.

### DIFF
--- a/src/prompt.c
+++ b/src/prompt.c
@@ -446,6 +446,7 @@ readline_init(void)
 
 static void sigint_absorb_handler(int sig) {
 	signal(SIGINT, SIG_DFL);
+	redraw_display(true);
 }
 
 char *


src/prompt.c
```diff
@@ -446,6 +446,7 @@ readline_init(void)
 
 static void sigint_absorb_handler(int sig) {
 	signal(SIGINT, SIG_DFL);
+	redraw_display(true);
```
jonas: It feels a bit unsafe to call this inside the signal handler. Mind changing it to use a flag?

0ax1: Good point. How about moving `redraw_display` out of `sigint_absorb_handler` and to the end of `read_prompt`? Based on your comment, I came to the conclusion that it probably shouldn't be in `sigint_absorb_handler` at all.

rolandwalker:
This was addressed in #618 for most cases.  There is still an issue at an empty prompt:

```
:^C
```

which might be related to accounting for `^C` having a visual width of two cells on screen.  (That is definitely an issue with quoted-insert <kbd>Ctrl-V</kbd> <kbd>Ctrl-A</kbd> in the middle of some regular text.)

0ax1:
@rolandwalker I always have this issue regardless whether the prompt is empty or not. 
I'm on macOS 10.12.5, tig `e709657`.

Do you see a way of not drawing `^C` to the prompt at all? Even when adding `redraw_display(true)` to the end of `read_prompt`, it is shown for a short moment before the display gets redrawn.

rolandwalker:
Terminals are weird - I'm on OS X too. Rendering the `^C` doesn't seem bad so long as it gets cleared.  The proposal to set a flag in the signal handler sounds sensible.

jonas:
I just pushed 2b384f6 which might also fix the prompt redraw issue mentioned in this PR. Please test.

0ax1:
Looks good. Thanks for fixing this!

PR #624, opened 2017-06-28
rfc: add `maxwidth` format param for fields `author`, `file-name`, `ref`

rolandwalker:
Rationale: certain columns are derived from data sources which are free and difficult to tame.  I regularly use a repo containing a commit with a mangled `author` field of length 575 characters.  Once merged into a shared repo, this is not practical to alter.  There are similar length issues with formalized branch names in large organizations.

The `width` parameter is rigid, and was somewhat misleadingly documented. `author,width=30` wastes screen when opening any other repo, whereas `author,maxwidth=30` is reasonable in all cases.

But I'm not sure about the interface proposed in this PR.  Perhaps you would rather
 * change `width` to behave as maximum width, following the existing documentation, or
 * have `maxwidth` available on all fields where `width` is possible, or
 * have `maxwidth` functionality implemented under the existing `width` param with something like `width=<10` or `width=-10`

### DIFF
--- a/doc/tigrc.5.adoc
+++ b/doc/tigrc.5.adoc
@@ -440,6 +440,8 @@ author::
 	- 'width' (int): Fixed width for the column. When set to a value between
 	  1 and 10, the author name will be abbreviated to the author's initials.
 	  When set to zero, the width is automatically sized to fit the content.
+	- 'maxwidth' (int): Maximum width of the column. Permit automatically
+	  sizing content, up to this limit.
 
 commit-title::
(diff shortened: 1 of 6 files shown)


jonas:
I have tried to recall why `width` was implemented that way. Probably because of how it works in CSS and also to eliminate the "pop-corn" effect in large repos when loading. What you are proposing definitely looks like an improvement so I will think a bit more about the options you have listed.

jonas:
Great idea. Looking forward to trying it out.

PR #625, opened 2017-06-29
make diff-highlight use ordinary configurable colors

rolandwalker:
`diff-add-highlight` and `diff-del-highlight` become ordinary colors independently governing the fg/bg/attributes of the highlighted cells.

These entries are added to the system tigrc such that the out-of-box configuration is unchanged, though if the user previously customized some diff colors, diff highlight cells will not follow them automatically.

Rationale: independent coloring of four states matches functionality of `diff-highlight` itself. Net reduction of logic.  Dimmer colors for the backgrounded regions are a readability improvement.

### DIFF
--- a/src/line.c
+++ b/src/line.c
@@ -69,12 +69,6 @@ get_line_info(const char *prefix, enum line_type type)
 	struct line_info *info;
 	struct line_rule *rule;
 
-	/* XXX: Map special highligh types to avoid special settings. */
-	if (type == LINE_DIFF_ADD_HIGHLIGHT)
-		type = LINE_DIFF_ADD;
-	else if (type == LINE_DIFF_DEL_HIGHLIGHT)
-		type = LINE_DIFF_DEL;
-
 	assert(type < line_rules);
 	rule = &line_rule[type];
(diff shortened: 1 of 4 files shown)


src/line.c
```diff
@@ -69,14 +69,13 @@ get_line_info(const char *prefix, enum line_type type)
...
-	if (type == LINE_DIFF_ADD_HIGHLIGHT)
-		type = LINE_DIFF_ADD;
-	else if (type == LINE_DIFF_DEL_HIGHLIGHT)
-		type = LINE_DIFF_DEL;
-
 	assert(type < line_rules);
 	rule = &line_rule[type];
+
+	/* special highlight types are not prefix-dependent */
+	if ((type == LINE_DIFF_ADD_HIGHLIGHT) || (type == LINE_DIFF_DEL_HIGHLIGHT))
```
jonas: I think it would make sense to allow `stage.diff-add-highlight` to use a different color in the stage view, but I don't feel strongly about that.

rolandwalker: Why not, when deleting code adds the feature.

jonas:
Thank you.

PR #626, opened 2017-06-29
add echo action

rolandwalker:
Untestable since `save-display` ignores the status area.

Usecase:

```
bind blame   O  @open '%(file)'
bind blob    O  @open '%(file)'
bind diff    O  @open '%(file)'
bind grep    O  @open '%(file)'
bind stage   O  @open '%(file)'
bind status  O  @open '%(file)'
bind tree    O  @open '%(file)'
bind help    O  :echo 'Nothing to open'
bind log     O  :echo 'Nothing to open'
bind main    O  :echo 'Nothing to open'
bind pager   O  :echo 'Nothing to open'
bind refs    O  :echo 'Nothing to open'
bind stash   O  :echo 'Nothing to open'
```

### DIFF
--- a/doc/manual.adoc
+++ b/doc/manual.adoc
@@ -487,6 +487,7 @@ Prompt
 |:script <file>		|Execute commands from `<file>`.
 |:exec <flags><args...> |Execute command using `<args>` with external
 			 user-defined command option flags defined in `<flags>`.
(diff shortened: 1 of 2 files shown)


jonas:
Makes sense. Thanks.

PR #629, opened 2017-06-29
raise titlebar count to highest known item on stop-loading action

rolandwalker:
After #622, the titlebar counts could be stuck lower than the maximum known value if the user did `stop-loading`.


src/tig.c
```diff
@@ -305,6 +305,7 @@ view_driver(struct view *view, enum request request)
 				report("Stopped loading the %s view", view->name),
 			end_update(view, true);
 		}
+		redraw_display(true);
```
jonas: Calling `update_title` would be better.

PR #630, opened 2017-06-30
preload last search from persistent history

rolandwalker:
Following the interface of `less`, preload the last search from the persistent history at init time, so that the query is immediately available to keystroke <kbd>n</kbd> (`find-next`).

Inherent bug: as discussed in #620, the command and search histories are merged: if the last entry in the history was a command, it will be treated wrongly as a search query.  But this is not different than pressing <kbd>&lt;Up&gt;</kbd> at the search prompt currently.

### DIFF
--- a/src/prompt.c
+++ b/src/prompt.c
@@ -525,12 +525,18 @@ prompt_teardown(void)
 void
 prompt_init(void)
 {
+	HIST_ENTRY *last_entry;
+
 	readline_init();
 	using_history();
 	stifle_history(HISTORY_SIZE);
 	read_history(prompt_histfile());
 	if (atexit(prompt_teardown))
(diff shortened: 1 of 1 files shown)


src/tig.c
```diff
@@ -728,6 +727,12 @@ main(int argc, const char *argv[])
 
 	while (view_driver(display[current_view], request)) {
 		view = display[current_view];
+
+		if (saved_search && *saved_search) {
+			string_copy(view->env->search, saved_search);
```
jonas: This can be moved out of the loop and copy directly to `argv_env`. To keep things simpler we can probably just do it in the readline version of `prompt_init`.

jonas:
Great, thanks for adding another great feature. I'll probably add an option to configure the number of history items.

PR #632, opened 2017-06-30
let Ctrl-C exit Y/N and menu prompts

rolandwalker:
followup to #619

### DIFF
--- a/src/prompt.c
+++ b/src/prompt.c
@@ -126,7 +126,7 @@ prompt_yesno_handler(struct input *input, struct key *key)
 
 	if (c == 'y' || c == 'Y')
 		return INPUT_STOP;
-	if (c == 'n' || c == 'N')
+	if (c == 'n' || c == 'N' || key_to_control(key) == 'C')
 		return INPUT_CANCEL;
 	return prompt_default_handler(input, key);
 }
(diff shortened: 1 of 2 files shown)


src/prompt.c
```diff
@@ -574,6 +574,11 @@ prompt_menu(const char *prompt, const struct menu_item *items, int *selected)
 			break;
 
 		default:
+			if (key.modifiers.control && (key.data.bytes[0] == 'C')) {
```
jonas: If I recall correctly regarding how the keys are encoded you are only supposed to read `data.bytes` if the multibytes flag is set. I think it would be safer to define a macro to match Ctrl-C that check the flags etc.

rolandwalker: But we are reading `data.bytes` two lines later at `if (items[i].hotkey == key.data.bytes[0]) {`.  Though perhaps the case should be `OK` rather than `default`.

I can't follow the logic behind the `multibytes` flag.  It is always set for control characters from `get_input()`, which is odd.  And the second condition here can never be true:

* https://github.com/jonas/tig/blob/e709657252f9fb09ff97c9cc6e2a159fc370ab27/src/display.c#L666

Was it originally meant to be `0x7F`? That would be logical as the alternative to `multibytes`.

jonas: The `multibytes` flag was added to support UTF-8 keybindings. Before Tig simply read the `int` returned by `wgetch`. `multibytes` is set if when ncurses returns one of its internal key codes.

> The curses wgetch() function
> will return any given one of these only if the corresponding k- capability
> is defined in your terminal's terminfo entry.

Maybe they no longer apply with the use of raw mode?

PR #633, opened 2017-06-30
pick up diff-highlight colors from git-colors

rolandwalker:
Should have been in #625, didn't realize it was this easy.


jonas:
Yes, good idea.

PR #634, opened 2017-06-30
don't read past end-of-string in quoted case of concat_argv

rolandwalker:
Incidentally remove unused `pos`.

This was corrupting some strings.  On OS X it's pretty easy to trigger by

```
bind generic n  ?echo "one two" three " " four
```

and then pressing "n" several times.


jonas:
Thanks and nice cleanup.

PR #635, opened 2017-07-01
improve check for all-space filename in diff stats

rolandwalker:
This also looks vulnerable to filenames containing `|` and should therefore use `strrchr()` as in `diff_common_read_diff_stat_part()`.

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -178,7 +178,7 @@ diff_common_add_diff_stat(struct view *view, const char *text, size_t offset)
 
 	/* Ensure that '|' is present and the file name part contains
 	 * non-space characters. */
-	if (!pipe || pipe == data || strcspn(data, " ") == 0)
+	if (!pipe || pipe == data)
 		return NULL;
 
 	/* Detect remaining part of a diff stat line:


src/diff.c
```diff
@@ -178,7 +178,7 @@ diff_common_add_diff_stat(struct view *view, const char *text, size_t offset)
 
 	/* Ensure that '|' is present and the file name part contains
 	 * non-space characters. */
-	if (!pipe || pipe == data || strcspn(data, " ") == 0)
+	if (!pipe || pipe == data || strlen(pipe) == len)
```
jonas: I'm trying to wrap my head around this one. Would be good to maybe extend the comment a bit with an example or provide a diffstat where this fixes a false positive.

I will work on making a column renderer which prints the line type. This would allow us to make this testable.

rolandwalker: I agree that it's hard to follow and will revise.

This doesn't fix a false positive, but was an attempt to make the code match the existing comment.

I noticed that the clause `strcspn(data, " ") == 0` can't ever be true, since we already scanned past leading space when initializing `char *data = start + strspn(start, " ")`.

rolandwalker:
What about this -- just deleting the clause?  `pipe == data` covers the case that `data` has nothing additional to `pipe`.

jonas:
Yes, sounds good.

PR #636, opened 2017-07-01
expand tig variables in echo action

rolandwalker:
Followup to #626.

I didn't understand the series of transformations in `exec_run_request()`

 * https://github.com/jonas/tig/blob/e709657252f9fb09ff97c9cc6e2a159fc370ab27/src/prompt.c#L972-L989

so I did something simpler.  And doesn't the `exec_run_request()` logic discard empty arguments?


jonas:
Some of the argv handling code discard empty arguments as a way to skip variables which expand to the empty string.

PR #637, opened 2017-07-01
enable scroll optimizations for Terminal.app and iTerm2

rolandwalker:
Tested for both optimizations.

Terminal.app is the Apple-supplied terminal emulator, and iTerm2 the most popular OSS alternative.  Both set environment variable `$TERM_PROGRAM`.


jonas:
Tested lightly. Good to have those tweaks enabled for these two macOS terminals.

Issue #419, opened 2015-06-12
Merge commit loses graphical "angle" if another branch starts from it


minaguib:
Currently looks like this:

```
2015-05-28 13:08 Mina Naguib             │ ● commit2 feat2
2015-05-28 10:47 Mina Naguib             │ ● commit1 feat2
2015-05-26 13:32 Mina Naguib             ●─╯ Merge branch 'feat1' into production
2015-05-25 16:56 Mina Naguib             │ ● [feat1] commit2 feat1
2015-05-25 16:52 Mina Naguib             │ ● commit1 feat1
```

I'd expect logically something like this:

```
2015-05-28 13:08 Mina Naguib             │ ● commit2 feat2
2015-05-28 10:47 Mina Naguib             │ ● commit1 feat2
2015-05-26 13:32 Mina Naguib             ●═╡ Merge branch 'feat1' into production
2015-05-25 16:56 Mina Naguib             │ ● [feat1] commit2 feat1
2015-05-25 16:52 Mina Naguib             │ ● commit1 feat1
```

SteveLTN:
I also noticed this problem. For me, it looks like
```
 M─┐
 │ o
 M─┘
 │ o
 M─┘
 │ o
 M─┘
```

quinox:
This one bit me today, I was about to scold a developer about not merging his branch :)

halfgaar:
I have the same issue. I was very confused by the graph, until I opened gitk.

zevv:
I too scolded a colleague for not merging, the end result was not pretty.

To make up I'm trying to fix this issue, any feedback appreciated:

https://github.com/zevv/tig/commit/984031cb09d26bffbdbc5471e87ac15b8a5964e7

PR #638, opened 2017-07-02
Fix graph display issues

swegener:
These commits fix a couple of graph display issues. Most issues have been found by running the graph drawing code on a large repo. The added regression tests are initially committed in wrong state and then fixed up to make it easier to visualize the issues they fix.

It also adds two new drawing types for merge commits that are also used as the parent for several other  commits.


jonas:
It looks like this fixes #419 and supersedes #605.

jonas:
Thank you @swegener. Awesome to see the graph rendering improved.

PR #639, opened 2017-07-02
Graph speedup

swegener:
The graph generation is currently a pretty strcmp() intensive process. By interning strings and replacing the strcmp() invocations by simple pointer comparisons, the graph generation can be sped up by a factor of
four in large repos.


swegener:
I've updated the commit with htab_empty(intern_string_htab);

jonas:
Great, thanks.

rolandwalker:
@swegener thanks this is gloriously faster.  This dwarfs the speedup from #622 but paradoxically gives #622 more impact, since the screen writes come faster.

PR #641, opened 2017-07-03
fix back/parent in tree-view: single pop and return

rolandwalker:
Immediately return `REQ_NONE` after pop on back/parent action in tree-view.  Otherwise back/parent will always perform two actions.

It looks like continuing on to `switch (line->type)` was the correct behavior prior to f893b0a3e.


jonas:
This indeed looks like a regression. Thanks for adding a test.

PR #642, opened 2017-07-03
Enable more symbolic keys: F13-F19, ShiftLeft/Right/Del/Home/End/Tab

rolandwalker:
Along with some doc cleanup.

Further additional symbolic keys are known to ncurses; this PR is limited to keys I happen to be able to generate under xterm and/or iTerm2.

`<ShiftLeft>` looks like a poor choice of interface at first glance, compared to something composable like `<Shift><Left>`.  But these keys are not actually composable, at least as filtered through ncurses.

Ncurses has a [finite list of supported shifted keys](https://github.com/mirror/ncurses/blob/b22573b1ba4b51da883fa5f805b52f153fa5fae9/include/Caps#L514-L542), the majority of which will not be available for modern keyboards/emulators.

That might be somewhat more clear to the user if the names were changed to *eg* `<ShiftedLeft>`.


doc/tigrc.5.adoc
```diff
@@ -582,14 +583,28 @@ bind main R		refresh
...
-Note that due to the way ncurses encodes `Ctrl` key mappings, `Ctrl-m` and
-`Ctrl-i` cannot be bound as they conflict with 'Enter' and 'Tab' respectively.
-Furthermore, ncurses does not allow to distinguish between `Ctrl-f` and
-`Ctrl-F`. Ncurses typically translates `Ctrl-<Space>` to `Ctrl-@`, which is
-available for binding. Finally, `Ctrl-z` is automatically used for process
-control and will suspend Tig and open a subshell (use `fg` to reenter Tig).
+Notes
+
+- Tig reads keystrokes via ncurses and is subject to various limitations. See
+  'ncurses(3x)' and 'terminfo(5)' (or termcap).
```
jonas: 👍

jonas:
Build failure is due to 2b384f69b45d00a9310d276fad2dfbfc0aabcc67.

PR #643, opened 2017-07-03
hide cursor unless at textual prompt

rolandwalker:


jonas:
Ended up manually merging this after testing. I think this makes sense.

PR #645, opened 2017-07-04
rfc: undocumented save-echo action for testing

rolandwalker:
Re: our conversation in #635, this was my notion for exposing more guts to the test suite: anything in an `%()` variable would be testable.

### DIFF
--- a/include/tig/io.h
+++ b/include/tig/io.h
@@ -96,6 +96,7 @@ enum status_code io_run_load(const char **argv, const char *separators,
 		io_read_fn read_property, void *data);
 char *io_memchr(struct buffer *buf, char *data, int c);
 
+enum status_code save_string(const char *path, const char *string);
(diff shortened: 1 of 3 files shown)


jonas:
Variables can already be tested with `assert-var`, e.g. in [`test/main/branch-var-test`](https://github.com/jonas/tig/blob/76b8810f5fb5c5d32585659397eb79c5e1b077f2/test/main/branch-var-test#L18)

```
:exec !assert-var "%(branch)" == "mp/feature"
```

Does this add additional functionality?

rolandwalker:
I didn't notice `assert-var`, which is a clever hack.  That approach seems better as it has no impact on C source.

Many more tests could be leveraging this.  My recent fake-chdir tests against full-screen contents are needlessly fragile compared to clean assertions on `%(directory)`.

PR #646, opened 2017-07-05
configurable truncation-delimiter w/ utf8 suggestion/special value

rolandwalker:
The `Midline Horizontal Ellipsis` character is exactly for this purpose.  No change to defaults, but the value becomes configurable.

I find the term "delimiter" confusing in this context, but followed the existing wording.

### DIFF
--- a/doc/tigrc.5.adoc
+++ b/doc/tigrc.5.adoc
@@ -190,6 +190,12 @@ The following variables can be set:
 
 	What type of character graphics for line drawing.
 
+'truncation-delimiter' (mixed) [utf-8|<string>]::
+
+	A single character to draw where columns are truncated. The default is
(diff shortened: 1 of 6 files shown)


jonas:
We can change the wording if you prefer. Do you prefer "ellipsis"?

rolandwalker:
"Ellipsis" over "truncation"?  Both read fine.

"Delimiter" is confusing compared to other choices such as "character", "symbol", "glyph", "marker", etc.  I suppose that's because delimiters **show** (or define) boundaries, and this mechanism is about **hiding**.

Just a point of English and not a big deal!

jonas:
Yeah, not being a native English speaker and not being much of a speaker at all makes for some confusing choice of words. ;)

PR #648, opened 2017-07-06
allow Ctrl-C to terminate file-finder prompt

rolandwalker:
This is a special case with its own input handler.  It's also a little annoying that <kbd>C-C</kbd> appears in the help screen for some cases and not for others.

Incidentally it is difficult to figure out that the `search` keymap applies to the file-finder rather than the `/` search.

### DIFF
--- a/src/ui.c
+++ b/src/ui.c
@@ -256,6 +256,12 @@ file_finder_input_handler(struct input *input, struct key *key)
 		file_finder_draw(finder);
 		return INPUT_SKIP;
 
+	case REQ_BACK:
+	case REQ_PARENT:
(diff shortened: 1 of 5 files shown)


jonas:
The `search` keymap should probably be renamed to `edit`? I initially introduced it when [working on](https://github.com/jonas/tig/tree/gh-5-incremental-search) incremental search (#5) and decided to use it for the file finder. An `edit` keymap makes more sense if we want to repurpose it for things like the `y/n` prompt etc.

jonas:
Maybe"input" could work. Will think about it.

PR #649, opened 2017-07-06
report_clear() efficiency, plus clear after grep

rolandwalker:
2b384f69b introduced a performance regression due to use of `wclear()`, which dirties the entire screen.  This can be seen in the form of flickering: enter the main view and hold down <kbd>&lt;Up&gt;</kbd>.

The core issue being addressed in 2b384f69b is that readline can leave an incorrect cursor position in virtual window `newscr` (which can be inspected using `getyx()`).

So, explicitly reset the `newscr` cursor position post-readline, which enables switching to the much more efficient `werase()` for erasing the status area.

Incidentally fix the missed case of `report_clear()` after a grep prompt.

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -430,9 +430,8 @@ report_clear(void)
 		return;
 
 	if (!input_mode && !status_empty) {
-		wclear(status_win);
-		wclrtoeol(status_win);
-		wnoutrefresh(status_win);
+		werase(status_win);
+		doupdate();
 	}
 	status_empty = true;
 	update_view_title(view);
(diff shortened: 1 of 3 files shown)


jonas:
2b384f6 was reverted and replaced with the report_clear code in aa84d85fef0eba5e02618e2070cf81d68f8b7bd4. I've made a follow up in #650 if you have time to review.

rolandwalker:
Sure, 2b384f6 onward.   aa84d85 has the same issue with calling `wclear()` frequently.

rolandwalker:
To clarify something I should have explained in the original comment here, `set_cursor_pos() … werase() … doupdate()` is a surgical strike that minimizes redrawing.

jonas:
Thanks for clarifying. I like the use of werase instead of wclear.

PR #650, opened 2017-07-06
Add special call to clear the status window after user input

jonas:
Follow-up to aa84d85fef0eba5e02618e2070cf81d68f8b7bd4 to fix performance
regression noted in #649. Also cleans up open_prompt to not clear the
status window.

### DIFF
--- a/include/tig/display.h
+++ b/include/tig/display.h
@@ -25,6 +25,7 @@ extern WINDOW *status_win;
 void update_status(const char *msg, ...);
 void report(const char *msg, ...) PRINTF_LIKE(1, 2);
 void report_clear(void);
+void status_clear(void);
 
 /*
  * Display management.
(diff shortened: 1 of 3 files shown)


rolandwalker:
This is nicer logic but doesn't address the performance regression in #649.

`set_cursor_pos(0, 0)` is the actual direct fix to the bug reported in #623.  Readline is the problem.  Ncurses fails to keep up with all the changes that may happen during readline world.  

`set_cursor_pos(0, 0)` forces ncurses to sync some state, and it follows readline because that is the locus of the bug.  `getyx(newscr …)` always returns incorrect values after Ctrl-C inside `readline()`.

`wclear()` also forces ncurses to sync state, just expensively, because `wclear()` marks the entire screen dirty for redrawing.   In this case we get a complete redraw on every keystroke.

PR #652, opened 2017-07-07
Distinguish between cancel action and empty string at all prompts

### LINKED ISSUE #627: tig jumps to a match after canceling search
How to reproduce:
- Search for sth. that has a match and press enter
- Start another search with '/' and cancel the search with C-C
- Instead of doing nothing, tig jumps to an instance of the previous match

rolandwalker:
Fixes #627
xrefs #583 #619 #623 #632 #648

The expectations for `/` search described in #627 match the behavior of `less`, which seems like a good conservative guide.

After this PR <kbd>Ctrl-C</kbd> means "cancel without action" at every prompt, which is accomplished by
 * maintaining the distinction between null pointers and strings of length zero, representing canceled input and empty input, respectively. This is somewhat fragile but also intuitive and consistent with most of the existing code paths.
 * at the input boundary, capturing the distinction by setting a flag in the readline signal handler as discussed in #623.
 * at the input boundary, when configured without readline, changing the way `read_prompt()` calls `read_prompt_incremental()`.  Note that <kbd>Ctrl-C</kbd> is unsupported at most prompts without readline; this merely compensates for the rest of the PR, and avoids regressions when compiled without readline

Since <kbd>Ctrl-C</kbd> has been made consistent, we may as well summarize here the behavior of an empty <kbd>&lt;Enter&gt;</kbd> across various prompts:

| prompt              | empty <kbd>&lt;Enter&gt;</kbd> | changed by this PR              |
| ------------------- |------------------------------- | ------------------------------- |
| `%(prompt)`         | accepted as empty string       | **yes**, previously threw error |
| `?<command>` (y/n)  | equivalent to `N` (No)         | no                              |
| `options`           | toggle current option          | no                              |
| `:`                 | cancel without action          | no                              |
| `/`                 | recall last search query       | no                              |
| `?`                 | recall last search query       | no                              |
| `grep-view`         | cancel without action          | no                              |
| file-finder         | choose file under cursor       | no                              |

Some might consider it poor UI that empty <kbd>&lt;Enter&gt;</kbd> has many different behaviors.  But in practice these are grounded in historical TUI behaviors, and all seem natural to me, with the exception of the Y/N prompt.  <kbd>&lt;Enter&gt;</kbd> should preferably not have a meaning at Y/N, and if it does, that should be indicated to the user *eg* `Question: y/N`.

### DIFF
--- a/src/prompt.c
+++ b/src/prompt.c
@@ -187,6 +187,8 @@ read_prompt_incremental(const char *prompt, bool edit_mode, bool allow_empty, in
 }
 
 #ifdef HAVE_READLINE
+static bool prompt_interrupted = false;
+
 static void
 readline_display(void)
 {
@@ -454,6 +456,7 @@ readline_init(void)
 
 static void sigint_absorb_handler(int sig) {
 	signal(SIGINT, SIG_DFL);
+	prompt_interrupted = true;
 }
 
 char *
@@ -479,10 +482,12 @@ read_prompt(const char *prompt)
 	/* readline can leave the virtual cursor out-of-place */
 	set_cursor_pos(0, 0);
 
-	if (line && !*line) {
+	if (prompt_interrupted) {
 		free(line);
 		line = NULL;
 	}
+
+	prompt_interrupted = false;
 
 	if (line)
 		add_history(line);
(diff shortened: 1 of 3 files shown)


src/prompt.c
```diff
@@ -187,6 +187,8 @@ read_prompt_incremental(const char *prompt, bool edit_mode, bool allow_empty, in
 }
 
 #ifdef HAVE_READLINE
+static bool prompt_interrupted = false;
```
jonas: I'll add volatile here.

PR #653, opened 2017-07-07
reduce flicker in file finder; avoid wclear()

rolandwalker:
It may be possible to optimize some details of `file_finder_draw()`, but this small change improves performance and reduces flicker by avoiding `wclear()` as in #649.

To visualize the flicker: run `tig | pv -qL5000` and hold the <kbd>&lt;Down&gt;</kbd> key.


jonas:
`wclear` considered harmful. Nice improvement. 👍

PR #654, opened 2017-07-07
permit diff cells of length 0, to reset diff-highlight background color

rolandwalker:
#625 introduced a regression: when `diff-highlight` is active, the trailing background color can run to the end of line.  Example: `tig show d4d0825ae`

This PR allows permits a cell of length zero at EOL which changes the color back.

I'm not certain that this won't have other side effects, nor that it is the best way to address the bug.

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -80,10 +80,10 @@ struct diff_stat_context {
 };
 
 static bool
-diff_common_add_cell(struct diff_stat_context *context, size_t length)
+diff_common_add_cell(struct diff_stat_context *context, size_t length, bool allow_empty)
 {
 	assert(ARRAY_SIZE(context->cell) > context->cells);
-	if (length == 0)
+	if (!allow_empty && (length == 0))
 		return true;
 	if (context->skip && !argv_appendn(&context->cell_text, context->text, length))
 		return false;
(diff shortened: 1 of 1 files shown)


src/diff.c
```diff
@@ -83,8 +83,6 @@ static bool
 diff_common_add_cell(struct diff_stat_context *context, size_t length)
 {
 	assert(ARRAY_SIZE(context->cell) > context->cells);
-	if (length == 0)
-		return true;
```
jonas: Would be great to only allow empty cells at the end of the line. Need to take a closer look.

jonas:
I am not able to reproduce this in iTerm and the default Terminal app.

rolandwalker:
It seems that `color diff-add-highlight` must be set.  I see the issue in any terminal, including xterm, with

 * checked out HEAD 8a71e6b62
 * `diff-highlight` on `$PATH`
 * the following command

```
GIT_CONFIG=/dev/null TIGRC_SYSTEM="$(pwd)/tigrc" TIGRC_USER=<(printf 'color diff-add-highlight black green\ncolor diff-del-highlight black red\nset diff-highlight = true\n') ./src/tig show d4d0825ae
```

diff-highlight-regression

rolandwalker:
Improved with an attempt to limit empty cells to EOL+line-contains-diff-highlight.  Still not sure I follow all of the relevant code.

jonas:
👍

PR #655, opened 2017-07-08
fix string corruption in file finder

rolandwalker:
~Allocate one more byte for file-finder items~ [edited]

Allocate the full `sizeof(struct file_finder_line)` rather than the size of a pointer, so that there is enough memory for the NUL terminator (and so that the NUL is pre-filled).

Otherwise `strncpy()` will take up the entire space, and tig may later read past the end.

### DIFF
--- a/src/ui.c
+++ b/src/ui.c
@@ -64,7 +64,7 @@ file_finder_read(struct file_finder *finder, const char *commit)
 			break;
 		}
 
-		finder->file[files] = calloc(1, sizeof(*finder->file) + buf.size);
+		finder->file[files] = calloc(1, sizeof(struct file_finder_line) + buf.size);
 		if (!finder->file[files]) {
 			ok = false;
 			break;


src/ui.c
```diff
@@ -64,7 +64,7 @@ file_finder_read(struct file_finder *finder, const char *commit)
 			break;
 		}
 
-		finder->file[files] = calloc(1, sizeof(*finder->file) + buf.size);
+		finder->file[files] = calloc(1, sizeof(*finder->file) + buf.size + 1);
```
jonas: Did you see this default? The `text` field in `file_finder_line` has size 1 to account for the NUL byte.

rolandwalker: Oh, I thought the first term was just to count the other member of the struct.

The issue is still there as described, because `sizeof(*finder->file)` returns constant `sizeof(uintptr_t)`, coincidentally `size_t`, the size of `matches`, the other member of the struct.

It should be `sizeof(**finder->file)` or `sizeof(struct file_finder_line)`.  The second one is better because it's harder to typo.  And either is better than my `+ 1` because of alignment.

jonas:
Thanks a lot for fixing this!

PR #658, opened 2017-07-10
clarify Shifted symbolic keys in news

rolandwalker:
Not sure it is important enough to list all of them, but if so it was the Shift* variants that were added.

F13-F19 is correct — not shifted.


jonas:
Thanks for the clarification.

jonas:
Hope to make a release with all your great changes soon.

PR #660, opened 2017-07-10
visually distinguish merges vs ordinary commits in utf8 graph

rolandwalker:
by returning smaller character "Bullet Operator" for ordinary commits


ericbn:
Even better, we should be able to configure the characters we want to use.

PR #662, opened 2017-07-12
let `set diff-highlight=true` work when the diff-highlight executable is not on $PATH

rolandwalker:
`diff-highlight` lives under contrib and is probably not on the user's `$PATH` in most installs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -311,6 +311,7 @@ TIG_OBJS = \
 	src/stash.o \
 	src/grep.o \
 	src/ui.o \
+	src/apps.o \
 	$(GRAPH_OBJS) \
 	$(COMPAT_OBJS)
 
--- a/configure.ac
+++ b/configure.ac
@@ -11,7 +11,7 @@ m4_pattern_forbid([^AX_])
 
 AC_PROG_CC
 
-AC_CHECK_HEADERS([execinfo.h stdint.h stdlib.h string.h sys/time.h unistd.h wordexp.h])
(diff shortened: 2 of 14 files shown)


jonas:
I played around with this one and I really like the idea how I'd prefer that the lookup does an `lstat` to verify the location of the diff-highlight script. WDYT?

rolandwalker:
Manually finding the location would also have the advantage of being able to invoke `diff-highlight` when the package manager failed to create the executable (at some revision it became required to run `make` in the `…/contrib/diff-highlight` dir).

rolandwalker:
Revised, with probably too much refactor.  Willing to squash as always.

* executing `git --exec-path` was relatively expensive so the result is preserved.  A few other things could realize similar savings.
* due to caching, setting `diff-highlight` multiple times as was done in the tests no longer works.  This is construed as a feature.
* logic about how to invoke the `diff-highlight` utility seems unrelated to viewing diffs, so it was spun out into `apps.c`.  If you don't like a new file, `options.c` would make more sense than `diff.c`.
* The "rescue" scenario started in recent versions of git, and happens to me currently with git from Macports.  The user lacks a shebanged `diff-highlight` executable unless an extra `make` step was done.  But this can be trivially worked around by invoking `perl` directly.  Unfortunately test coverage for this code path doesn't seem possible at the moment.

Handling of`diff-highlight` execution failure wants a few improvements, but was left unchanged.

rolandwalker:
valgrind doesn't like the final test case

rolandwalker:
valgrind issue addressed

PR #664, opened 2017-07-14
add a TODO test category

rolandwalker:
It would be more fancy to run TODOs every time, and report unexpected successes.

Instead this PR creates a `TODO` category which is a simple subset of `skipped`.  TODOs can be executed by adding the string `todos` to `$TEST_OPTS`.

Individual selections within `run_test_cases` are supported, ~~but it would be a bit more work to make the final summary statement have a correct count of skipped TODOs from `run_test_cases`.~~

Edit: updated to 
 * count `run_test_cases` TODOs correctly in `show-results.sh`
 * include an example usage: a test which currently exercises a segfault

### DIFF
--- a/test/tigrc/segfault-test
+++ b/test/tigrc/segfault-test
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+. libtest.sh
+
+test_todo "Segfault without 'set reference-format' in tigrc"
+
+export TIGRC_SYSTEM=/dev/null
+
+tigrc <<EOF
+bind generic : prompt
+# sufficient to rescue from crash
+# set reference-format = branch
+EOF
+
+steps ':save-options'
+
+test_tig status
+
+assert_equals stderr <<EOF
+EOF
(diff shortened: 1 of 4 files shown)


rolandwalker:
Picked up a missing backslash from #616.

test/tigrc/segfault-test
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+. libtest.sh
+
+test_todo "Segfault without 'set reference-format' in tigrc"
```
jonas: Is this a test for the todo marker or is it something you plan to fix in a future PR?

rolandwalker: It's a real tig bug, though I haven't looked into the cause.

My motivation is that TODO test PRs are more substantive than issue discussions.

jonas: Yes,  that is true, being able to add a reproducer and track its status.

Issue #232, opened 2014-01-24
Unexpected author abbreviations when restoring from background


stuarthicks:
Using tig 1.2.1 (Nov 22 2013) on OSX 10.8.5, iTerm 2 and zsh.

Steps.
1. Launch tig
2. Press Up arrow (message appears on screen to say you cannot move above the top row)
3. Press ctrl-z to suspend tig
4. Bring tig back to the foreground with `fg`
5. Press Up arrow again

Expected: To see message on screen again about being at the top of the list.
Actual: Author field becomes abbreviated.

This can be repeated to un-abbreviate the author field. So I presume calling a toggleAuthorAbbreviation function somehow.

Also observed, immediately after restoring from background, I have to press the down arrow twice to go down. Are key mappings offset/incorrect in this state?

igoralmeida:
I think a similar problem happens if you suspend $EDITOR while editing a commit message. Only tested with vim+bash:

1. In the status view, have some changes ready and press `C` to start a commit
1. Press `control-z` to suspend, shell shows up. Run `fg` to get vim back to foreground.
1. Finish the commit message, save and quit. Commit output is shown, press enter and tig shows up again.
1. Screen is not updated, it seems: keys pressed are just printed to the statusbar. I type `q<CR>` as many times as necessary to quit tig at this point.

commonquail:
I can't reproduce @stuarthicks' issue on 4b0aba8 ("erase status line at exit time") but I traced @igoralmeida's issue to 83df6c3 ("Remove unneeded calls to {def,reset}_prog_mode"). It's consistently reproducible and it's not necessary write or save anything. Reverting that commit restores editor suspension capability, however, when eventually quitting tig the terminal screen doesn't clear like it normally does (and like `top` vs `htop`); same behaviour as in tig 2.0.1.

jonas:
@commonquail Thanks for tracking down the commits. I reproduced @igoralmeida 's issue with the latest version and managed to come up with #657 to fix that.

jonas:
Helped by @rolandwalker 's work in #665 and 167d59ed1474e10ec3310fe7aa77314247332954 I believe I finally found a fix for the initial issue also.

stuarthicks:
@jonas @rolandwalker Thanks! 😄

PR #657, opened 2017-07-10
Fix suspension capability when calling out to external commands

jonas:
Clear the terminal before calling out to an external command so the
altscreen won't mess up any downstream subshell management. Restore
terminal settings as per init_display when restoring. Reproducible with
example from[0]:

 - Run `tig status`
 - Press 'e' to open Vim
 - Press Ctrl-Z to suspend Vim
 - Run `fg`
 - Quit Vim

 [0] https://github.com/jonas/tig/issues/232#issuecomment-313790438

Refs: #232
cc: @commonquail

PR #665, opened 2017-07-14
flush buffered keystrokes after external viewer

rolandwalker:
by ~~reopening the TTY~~ seeking to `SEEK_END`.  Otherwise keystrokes pressed prior to <kbd>Enter</kbd> hang around and can get read later.

To reproduce:
 * `:bind generic o !ls`
 * <kbd>o</kbd> # *shows listing*
 * <kbd>q</kbd><kbd>Enter</kbd> # *leaves `q` unflushed*
 * <kbd>o</kbd> # *does not show listing; returns immediately due to unflushed character*

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -64,6 +64,7 @@ open_external_viewer(const char *argv[], const char *dir, bool silent, bool conf
 				fprintf(stderr, "%s", notice);
 			fprintf(stderr, "Press Enter to continue");
 			getc(opt_tty);
+			fseek(opt_tty, 0, SEEK_END);
 		}
 	}
 


jonas:
Looks like this will fix #232

jonas:
I misread, but hope to have found a fix for #232 because of this ticket.

src/display.c
```diff
@@ -64,6 +64,9 @@ open_external_viewer(const char *argv[], const char *dir, bool silent, bool conf
 				fprintf(stderr, "%s", notice);
 			fprintf(stderr, "Press Enter to continue");
 			getc(opt_tty);
+			opt_tty = freopen(NULL, "r", opt_tty);
```
jonas: If I read `freopen(3)` the mode should be `"w+"` to match how it is initially opened in pager mode: https://github.com/jonas/tig/blob/1f757f6579f6c56b68bbd68c6896a3c1f204b859/src/display.c#L557

jonas:
Looks good to me, but I am not sure if this means that the terminal also needs to be reopened somehow. From my understanding of how `newterm` works the `FILE *` passed there is used for setting up the terminfo DB etc.

jonas:
Also somewhat related to the suspending cleanups in #657

jonas:
Yes, that looks safer.

PR #666, opened 2017-07-14
let external commands be much longer than SIZEOF_STR

rolandwalker:


jonas:
Curious what type of commands you are running. ;-)

rolandwalker:
Naturally I want to hit the GitHub API and display the result in the echo area.

PR #668, opened 2017-07-14
doc linting

rolandwalker:
We can squash these later but I think the similar #616 suffered from integrating too much.

### DIFF
--- a/doc/tigrc.5.adoc
+++ b/doc/tigrc.5.adoc
@@ -557,6 +557,7 @@ built-in keybindings.
 
 Keybindings at the line-entry prompt are typically governed by the readline
 library, and are configured separately in `~/.inputrc`.  See 'readline(1)'.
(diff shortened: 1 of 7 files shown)


doc/tigrc.5.adoc
```diff
@@ -803,7 +808,7 @@ View manipulation
 |quit                    |Close all views and quit
 |=============================================================================
 
-View specific actions
+View-specific actions
```
jonas: Do you mind to also update the `REQ_GROUP("View specific actions")` in [`include/tig/request.h`](https://github.com/jonas/tig/blob/2f55f1dd0635c99ebc9d5640b149239e52e0c26d/include/tig/request.h#L42)?

There's a `make update-docs` target which extracts documentation from the code.

jonas:
Thanks a lot for reviewing the docs.

jonas:
Thanks, will create a merge.

PR #669, opened 2017-07-14
Fix #232: Do not generate any input when handling Ctrl-Z

### LINKED ISSUE #232: Unexpected author abbreviations when restoring from background

jonas:
Move the special case of Ctrl-Z introduced in
167d59ed1474e10ec3310fe7aa77314247332954 so no input is generated by
get_input.

---

@rolandwalker does this seem reasonable to you?

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -43,6 +43,7 @@ Bug fixes:
  - Set cursor position when not updating prompt contents. (GH #647)
  - Erase status line at exit time for users without altscreen-capable terminals.
    (GH #589)
+ - Fix unexpected keys when restoring from suspend (`Ctrl-Z`). (GH #232)
  - contrib/vim.tigrc: Also bind G in the main as a workaround for limitations of
    the `none` action. (GH #594, #599)
  - Only override `blame-options` when commands are given and fix parsing of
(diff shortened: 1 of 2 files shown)


rolandwalker:
Seems reasonable, though the interplay here is difficult to work through.

It is odd to think that #232 couldn't be solved if tig was still using `cbreak()` instead of raw mode.

jonas:
It probably could have been solved, but I couldn't find a way to detect and flush the TTY input stream. You change to use raw mode made a fix more obvious.

PR #670, opened 2017-07-15
whitespace linting

rolandwalker:
A general whitespace cleanup offered as an apology for bungling the indents in `test/tigrc/width-test`.  Cleanup tends to reduce merge conflicts over the long term though it usually creates some in the short term.

 * tabs instead of spaces to indent
 * tabs instead of spaces to delimit columns
 * rm spaces interior to tabs
 * rm trailing space
 * rm stray linefeeds
 * add missing indents


jonas:
OK, let the rebasing begin.

PR #671, opened 2017-07-15
allow overriding sed executable in makefile

rolandwalker:
because some of the existing commands require GNU sed.  In particular, I needed to

```bash
make update-docs SED=gsed
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,11 @@ DOCBOOK2PDF ?= docbook2pdf
 
 LCOV ?= lcov
 GENHTML ?= genhtml
+ifneq (,$(shell which gsed))
+SED ?= gsed
+else
+SED ?= sed
+endif
 
 all: $(EXE) $(TOOLS)
 all-debug: all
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -74,6 +74,7 @@ DOCBOOK2PDF ?= docbook2pdf
 
 LCOV ?= lcov
 GENHTML ?= genhtml
+SED ?= sed
```
jonas: Given that all scripts assume GNU sed, I think it is safer to default to `gsed`.

jonas: OK, looks like this was a bad idea. I'll reenable my `sed -> gsed` alias and maybe configure a build that runs on macOS to ensure that the portability fixes you did in the test suite won't regress.

rolandwalker:
Sec, it shouldn't be hard to check for both and prefer `gsed`, since GNU make is already needed for this Makefile.

jonas:
Sweet!

PR #672, opened 2017-07-15
make $TEST_OPTS respect wordsplit under bash

rolandwalker:
Options were ignored when more than one was present:

```
TEST_OPTS="verbose todos" make test 
```

since IFS was set conditionally https://github.com/jonas/tig/blob/5d86622e4b9c29f6b948c4e463de7b026d4f9119/test/tools/libtest.sh#L18-L21 .

In the refactor, you might not approve renaming `$TIG_TEST_OPTS` to `$MAKE_TEST_OPTS`.  My reasoning was that I kept forgetting which one I was meant to set at CLI.

 * force wordsplit by `$IFS`
 * `for` loop is easier than `set`/`shift`
 * doc space-separation not comma
 * accept some user typos
 * die on unknown options
 * stop Makefile from setting `$TIG_TEST_OPTS` to `@` (which was default unless `make test V=1`)
 * recast `$TIG_TEST_OPTS` as `$MAKE_TEST_OPTS`
 * always set `$MAKE_TEST_OPTS` in Makefile; user interface is `TEST_OPTS`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,7 +214,11 @@ clean-test:
 test: clean-test $(TESTS)
 	$(QUIET_SUMMARY)test/tools/show-results.sh
 
-export TIG_TEST_OPTS = $(V:1=no-indent)
+ifneq (,$(strip $(V:@=)))
+export MAKE_TEST_OPTS = no-indent
+else
+export MAKE_TEST_OPTS =
+endif
 
 $(TESTS): PATH := $(CURDIR)/test/tools:$(CURDIR)/src:$(PATH)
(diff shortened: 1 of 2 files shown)


jonas:
LGTM, regarding

> In the refactor, you might not approve renaming $TIG_TEST_OPTS to $MAKE_TEST_OPTS. My reasoning was that I kept forgetting which one I was meant to set at CLI.

The `TIG_` prefix is used for defaults variables which can be overridden, one example is TIG_NCURSES. I don't have a strong opinion about it and from the `libtest.sh` point-of-view the name makes sense.

Issue #522, opened 2016-09-02
date disappears if main-view-date-format string is too long


alexch:
My original goal was to add the time zone to the date column, like this (in `~/.tigrc`):

```
set main-view-date = custom
set main-view-date-local = true
set main-view-date-format = "%F %R %Z"
```

but that made the entire column disappear. On further investigation, it seems that if the `main-view-date-format` result string is longer than 16 characters, the whole result is empty and the column shrinks. e.g. these all display fine:

```
set main-view-date-format = "%R %Z"

set main-view-date-format = "%F %R"

set main-view-date-format = "%Z %Z %Z"
```

This happens regardless of the `main-view-date-width` value.

---

...and I think I just found another bug - - if `set main-view-date-local = false` then zoned times inside the commit log are _not_ shifted to UTC, but the zone is displayed as UTC:

e.g. a commit date of 

```
Wed Aug 31 09:03:00 2016 -0700
```

is displayed as

```
09:03 UTC
```

but it should be either

```
16:03 UTC
```

or

```
09:03 PDT
```

I'd be happy to file a separate issue for that if you like.

jonas:
Thanks for the detailed report. The first issue is trivial to fix, however the time zone formatting is a bit harder. I've only managed to fix it so that time zone offsets can be shown. I think this is in line with how Git displays date, since Git only stores time zone offsets in the commit dates.

PR #673, opened 2017-07-15
Custom date format fixes

jonas:
Fix #522 by increase size of the internal date buffer and support time zone offsets.


jonas:
Superseded by #811

PR #674, opened 2017-07-15
wordexp() support and compat fallback

rolandwalker:
In #168 you expressed concern about platforms without `wordexp()`.  Without being sure what those platforms would be, I included a `compat_wordexp()` that can only expand `~/`.

I think the `compat_wordexp()` here is OK so far as it goes, but `expand_path()` is clunky at best.

`wordexp()` is a standard, but it's a lousy standard.  It does several types of transformation, such as arithmetic expansion, and exposes an off switch for only one type (`WRDE_NOCMD` for command expansion).  So `expand_path()` constrains `wordexp()` by constraining its inputs and rebuilding the result.

All that's gained is the rarely-used `~username` expansion, and the ability to use the word "standard".  Though it would be nice to have expansion in more places, like `bind generic * :script ~/.tig/scripts/star.command`.

~Note: this PR makes no functional change, and merely exposes `expand_path()`.  I was doubtful that `io_open()` should always expand its arguments.~

Edit: routed existing `:source` expansion through `expand_path`, with test.  Added `:script` path expansion, which doesn't seem testable since tests are already driven by script.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -132,6 +132,41 @@ get_path_encoding(const char *path, struct encoding *default_encoding)
 }
 
 /*
+ * Path manipulation.
+ */
+
+bool
+expand_path(char *dst, size_t dstlen, const char *src)
+{
+	if (!src)
+		return false;
+
+	if (src[0] == '~') {
+		/* constrain wordexp to tilde expansion only */
+		const char *ifs = getenv("IFS") ? getenv("IFS") : " \t\n";
+		wordexp_t we_result;
+		size_t metachar_pos;
+		char metachars[SIZEOF_STR];
+		char leading[SIZEOF_STR];
+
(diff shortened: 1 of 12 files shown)


src/io.c
```diff
@@ -135,6 +135,39 @@ get_path_encoding(const char *path, struct encoding *default_encoding)
  * Executing external commands.
  */
 
+bool
+expand_path(char *dst, const char *src)
```
jonas: Would be good to pass a `size_t dstlen` to keep things safer.

jonas:
LGTM. `io_open` already has a hack to open stdin, and I prefer to not add more. Anyway, I think `expand_path()` should be opt-in. It would also be good to have at least one use case for `expand_path()` in the PR.

jonas:
Thanks, a nice improvement.

Issue #557, opened 2016-12-20
Print external command output in the status bar


simark:
Is it possible to have an external command print something in the bottom status bar?  For example, I am thinking of enhancing the "copy sha1 to clipboard" example to have it print something to let the user know that the action was triggered:

    bind generic 9 !sh -c "echo -n %(commit) | xclip -selection c && echo Copied %(commit) to clipboard"

However, instead of hiding the tig interface and show the output along with "Press Enter to continue", it would be nice to show this in the status bar.  Is there any way to do this?

jonas:
Fixed as of commit 4a3d90a

Issue #200, opened 2013-09-18
Empty or one-line output from external commands should not create "Press ENTER" prompt


majutsushi:
I would like to echo a short status message after the suggested "copy revision to clipboard" binding to confirm that the command has been executed, but that requires me to hit Enter after every copy which is quite annoying. If the output is short enough to fit into one line there is no real need to prompt the user for confirmation.

linduxed:
I'm wondering the same thing. Is there some way to echo out something to the status line at the bottom?

edi9999:
I like that idea too

jonas:
Me too ;).

But I would prefer it would require a special flag(suggestions are welcome)
that turns this on since capturing output could increase memory usage
considerably for verbose commands. Ideally this flag would cause the
command to be executed with stdout being a pipe from which tig can read and
simply remember the last line.

edi9999:
Maybe something like :

`bind generic 9 <@ sh -c 'echo "copied"'`

The `<` meaning take the output from that command to tig.

edi9999:
I think that's how it works also with the confirm flag with `?@`

edi9999:
Any update on this ?

jonas:
Didn't get around to working on this, yet.

On Mon, Jun 22, 2015 at 7:48 AM, Edgar Hipp <notifications@github.com>
wrote:

> Any update on this ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jonas/tig/issues/200#issuecomment-114076420>.
>



-- 
Jonas Fonseca

jonhiggs:
I'd like to +1 this issue too. My first guess to fix it was adding a <CR> or two to the end but that didn't work.

nishantvarma:
This can also help when you do a git push origin right? Would like to avoid the extra enter usually  ... the alternative currently is to do pull in Vim which can execute a command silently.

netei:
Yes, if that syntax was implemented, you could do something like : 

    bind generic p <@ sh -c 'git push origin master > /dev/null && echo "Done" || echo "Fail"'

jonas:
Fixed in 4a3d90a, which uses the `+` modifier. I am considering to rewire it to `<` (which currently means quit-after-command).

PR #678, opened 2017-07-18
single-line output of external command to status bar

rolandwalker:
Adds new external command option flag `+`, meaning "echo first line of the output to the status bar".  Unlike `!`, output is integrated into the regular TUI and need not be manually dismissed.

Example usage:

```ini
# Show commit statistics for the author under the cursor
bind main U +sh -c 'git --no-pager shortlog -s --author="$(git show -s --format=%aE %(commit))" </dev/tty'
```

I think this is a good feature, but recognize that `open_external_viewer()` is taking too many params.

It might be better if the feature was "collapse whitespace and display as much output as fits in one screen width", though the user can get that by appending `| xargs` to a command.

(Odd little bug in `git-shortlog`: no output from the above command without `</dev/tty`.)

Edit: flagname `+`

### DIFF
--- a/doc/tigrc.5.adoc
+++ b/doc/tigrc.5.adoc
@@ -640,6 +640,8 @@ the command that should be executed.
 |=============================================================================
 |!			|Run the command in the foreground with output shown.
 |@			|Run the command in the background with no output.
+|+			|Run the command synchronously, and echo the first line
+			 of output to the status bar.
 |?			|Prompt the user before executing the command.
 |<			|Exit Tig after executing the command.
 |=============================================================================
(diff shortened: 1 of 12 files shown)


jonas:
Very nice. Didn't look yet, but from the description I'd prefer another symbol than `_` because it is hard to read. Unfortunately `<` is used for quitting so maybe `>`?

Seems to fix #557 and #200.

rolandwalker:
I actually wanted `>` for something else.  Would `%` be okay?  Or `+`?

jonas:
Yes, `+` looks better. Although I am almost up for using `<` and finding another modifier for exiting.

PR #683, opened 2017-07-22
WIP improved context-sensitive view-blob

rolandwalker:
I would like to improve the sensitivity of some tig actions to their context.

This is expressed here as three TODO tests which call `view-blob` from the diff view.

Tig already adopts the model of context-sensitive action here, but it depends on the cursor directly touching a line which it can associate to a blob.  When that is not true, tig opens the last blob it can remember, or opens the file finder.

Neither of those is likely to match user expectations.  But when viewing a diff, the cursor is always either **on** a blob-associated line, or **before** some blob-associated line.  So I propose that the simplest intuitive context-sensitive `view-blob` is
 * view this line's blob, or failing that
 * scan forward by lines and view the next associated blob

### DIFF
--- a/include/tig/diff.h
+++ b/include/tig/diff.h
@@ -40,7 +40,7 @@ enum status_code diff_init_highlight(struct view *view, struct diff_state *state
 bool diff_done_highlight(struct diff_state *state);
 
 unsigned int diff_get_lineno(struct view *view, struct line *line);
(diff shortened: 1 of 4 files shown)


jonas:
This makes sense and sounds like it will bring `view-blob` more in line with how the edit action works. However it is a bit hard to understand the behavior from only looking at the todo tests.

PR #684, opened 2017-07-22
todo test for bug re: jumping past last line

rolandwalker:
`:0` is also possible and somewhat debatable.


jonas:
Looks easy enough to fix.

PR #686, opened 2017-07-26
don't segfault on :exec irregular arguments

rolandwalker:
Bug: segfaults on `:exec` (no args) and `:exec !` (flags but no actual command args).

Created a directory `test/prompt/` here, which is debatable, since the included tests don't work by stuffing keystrokes.  But it would be a good thing for the test suite to learn how to stuff keystokes, and populate a proper `test/prompt/`.


jonas:
What do you think it takes to get those tests working?

rolandwalker:
this one does include the fix

PR #688, opened 2017-07-26
timeouts on external processes in test harness

rolandwalker:
Timeouts are a prerequisite for tests to be able to drive tig by simulating keystrokes.

Core sections of `run_test_cases` and `test_tig` simplified via use of subshells.

Timeouts disabled under valgrind since this combination is so far untested.

### DIFF
--- a/test/tools/libtest.sh
+++ b/test/tools/libtest.sh
@@ -204,6 +204,7 @@ debugger=
 trace=
 todos=
 valgrind=
+timeout=10
 
 ORIG_IFS="$IFS"
 IFS=" 	"
@@ -216,6 +217,7 @@ for arg in ${MAKE_TEST_OPTS:-} ${TEST_OPTS:-}; do
 		no[-_]indent|noindent) indent= ;;
 		debugger=*) debugger="$(expr "$arg" : 'debugger=\(.*\)')" ;;
 		debugger) debugger="$(auto_detect_debugger)" ;;
+		timeout=*) timeout="$(expr "$arg" : 'timeout=\(.*\)')" ;;
 		trace) trace=yes ;;
 		todo|todos) todos=yes ;;
 		valgrind) valgrind="$HOME/valgrind.log" ;;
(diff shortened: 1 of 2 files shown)


jonas:
I got a bit behind on your stream of patches so this one needs to be rebased

rolandwalker:
Given timeouts, libtest can simulate keystrokes with

```bash
python -c "import fcntl,os,termios,tty; fd = os.open('/dev/tty',os.O_RDWR); tty.setcbreak(fd); [fcntl.ioctl(fd,termios.TIOCSTI,ch) for ch in '$KEYSTROKES']"
```

test/tools/libtest.sh
```diff
@@ -468,42 +479,44 @@ test_tig()
...
+			runner=exec
+			# FIXME: Tell Valgrind to forward status code
+			if [ "$expected_status_code" = 0 ] && [ -n "$valgrind" ]; then
+				runner=valgrind_exec
+				timeout=0
+			fi
+			if [ -s "$HOME/${prefix}stdin" ]; then
+				exec 4<"$HOME/${prefix}stdin"
+			elif [ -t 0 ] && [ -e /dev/tty ]; then
+				exec 4</dev/tty
```
jonas: Since c883298355e02fc2c431cb68fe5edc212ec62fea this is not necessary since Tig opens `/dev/tty` by default.

test/tools/libtest.sh
```diff
@@ -468,42 +479,44 @@ test_tig()
...
 			else
-				(cd "$work_dir" && tig "$@") > "${prefix}stdout" 2> "${prefix}stderr.orig"
+				exec 4<&0
 			fi
+			"$runner" tig "$@" <&4 > "$HOME/${prefix}stdout" 2> "$HOME/${prefix}stderr.orig" &
+			tig_pid="$!"
+			install_pid_timeout "$tig_pid"
```
jonas: One requirement for merging this is that timeout failures are clearly identified in the test output. For example
```
[FAIL] Test timed out after $timeout seconds
```
Is that the case?

rolandwalker: That's not the case, and it is a little grotty with straight POSIX, but should be doable given another subshell.

PR #689, opened 2017-07-26
untruncate preloaded search string

rolandwalker:
`string_copy()` respects `sizeof()` on its source rather than `strlen()`, truncating the preloaded search string to the sizeof a pointer.

This fixes it, though I actually don't understand why `string_copy()` wouldn't consult `strlen()`.

### DIFF
--- a/src/prompt.c
+++ b/src/prompt.c
@@ -590,7 +590,7 @@ prompt_init(void)
 
 	last_entry = history_get(history_length);
 	if (last_entry)
-		string_copy(argv_env.search, last_entry->line);
+		string_ncopy(argv_env.search, last_entry->line, strlen(last_entry->line));
 }
 #else
 char *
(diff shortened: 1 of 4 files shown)


jonas:
I am OK with changing `string_copy`. It was made that way since a lot of places copy one static buffer to another so it was simpler to not call `strlen()` and instead truncate based on `NUL` in the source buffer.

jonas:
Not sure I understand how this PR relates to #703

rolandwalker:
Not related to other testing PRs.

It was necessary here to add `test_requires readline` to pass Travis.  One Travis run is done based on a simple `contrib/config.make` which doesn't have readline support.  Thus a test of whether readline recalls the previous search could not pass.

Even so the `test_requires readline` here is not ideal.  A `tig --version --verbose` that exposes compile info would be nicer.

PR #691, opened 2017-07-27
incorporate xterm wcwidth.c to compute unicode widths

rolandwalker:
It turns out that #679 can be improved upon by incorporating a single file directly from xterm.  This is a fork of the same code with numerous tweaks.  The single-file format would make it easier to track Unicode revisions by simply tracking xterm.  Coincidentally by the same author as ncurses.


jonas:
Awesome

PR #692, opened 2017-07-27
valgrind test suite fixes

rolandwalker:
* correctly find per-kernel suppressions file
* must set IFS to split `$valgrind_ops`
* move suppressions to separate variable to solve quoting issues
* forward tig exit status

### DIFF
--- a/test/tools/libtest.sh
+++ b/test/tools/libtest.sh
@@ -438,23 +438,31 @@ test_setup()
 valgrind_exec()
 {
 	kernel="$(uname -s 2>/dev/null || printf 'unknown\n')"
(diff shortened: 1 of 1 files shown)


test/tools/libtest.sh
```diff
@@ -438,23 +438,31 @@ test_setup()
 valgrind_exec()
 {
 	kernel="$(uname -s 2>/dev/null || printf 'unknown\n')"
-	kernel_supp="test/tools/valgrind-$kernel.supp"
+	kernel_supp="$HOME/../../../tools/valgrind-$kernel.supp"
```
jonas: This might be better defined as `$base_dir/test/tools/valgrind-$kernel.supp`

PR #693, opened 2017-07-28
fix insufficient memmove in compat/wordexp.c

rolandwalker:
include the NUL byte in the first `memmove`.

### DIFF
--- a/compat/wordexp.c
+++ b/compat/wordexp.c
@@ -19,6 +19,7 @@
 #include "stddef.h"
 #include <stdlib.h>
 #include <string.h>
+#include <stdio.h>
 
 void
 compat_wordfree (wordexp_t *pwordexp)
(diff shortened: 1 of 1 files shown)


compat/wordexp.c
```diff
@@ -40,7 +40,7 @@ compat_wordexp (const char *words, wordexp_t *pwordexp, int flags)
 		pwordexp->we_wordv[0] = realloc(pwordexp->we_wordv[0],
 						strlen(pwordexp->we_wordv[0]) + strlen(home));
 		memmove(pwordexp->we_wordv[0] + strlen(home) - 1, pwordexp->we_wordv[0],
-			strlen(pwordexp->we_wordv[0]));
+			strlen(pwordexp->we_wordv[0]) + 1);
```
jonas: How about something like this to avoid `memmove` and improve error checking.
```c
int
compat_wordexp (const char *words, wordexp_t *pwordexp, int flags)
{
	char *expanded = NULL;

	if (words[0] == '~' && (words[1] == '/' || words[1] == 0)) {
		const char *home = getenv("HOME");
size_t len = strlen(home) + strlen(words + 1) + 1;

		if (home) {
			size_t len = strlen(home) + strlen(words + 1) + 1;

			if ((expanded = malloc(len)) && !string_nformat(expanded, len, "%s%s", home, words + 1) {
				free(expanded);
				return -1;
			}
		}
	} else {
		expanded = strdup(words);
	}

	if (!expanded)
		return -1;

	pwordexp->we_wordv = calloc(2, sizeof(*pwordexp->we_wordv));
	if (!pwordexp->we_wordv) {
		free(expanded);
		return -1;
	}
	pwordexp->we_wordv[0] = expanded;

	return 0;
}
```

jonas:
Great thanks.

PR #697, opened 2017-07-28
save-options able to write empty reference_format

rolandwalker:
which happens if attempting to write out the builtin config.  Fixes segfault todo from #664.


jonas:
Sweet. Good to have this todo test resolved.

PR #698, opened 2017-07-28
tilde expansion was added to :script

rolandwalker:
very slightly improved in `:source` but not news-worthy


jonas:
Yes, good point

PR #699, opened 2017-07-28
expose default save-action files to assert_equals

rolandwalker:
by bumping them up to test `$HOME`, so

```
steps ':save-options'
…
assert_equals tig-options.txt <<EOF
option content
EOF
```


jonas:
My gut feeling is that it is better to explicitly specify the output file so it is clear what is being asserted. But this brings up a good point that we don't replace the path of `:save-options` inside `tig_script`.

PR #700, opened 2017-07-28
accept make /full/path/to/test-test

rolandwalker:
Pro: doesn't change the default case.

Con: I am probably the only user.


rolandwalker:
#688 caused, or flushed out, a timeout problem here after rebase https://travis-ci.org/jonas/tig/jobs/263825104

jonas:
I prefer to keep paths relative.

PR #701, opened 2017-07-29
tolerate empty lines in scripts

rolandwalker:
The entire test suite provides coverage after removing a sed directive.


jonas:
Thanks

PR #702, opened 2017-07-29
readline-complete on repo variables

rolandwalker:


jonas:
Oops, looks like an oversight. Thanks for fixing.

PR #703, opened 2017-07-29
sidestep readline when executing a script

rolandwalker:
Readline was causing major perf issues when interpreting a script, and inappropriately storing each script line in `~/.tig_history`.

### DIFF
--- a/src/prompt.c
+++ b/src/prompt.c
@@ -127,6 +127,20 @@ prompt_default_handler(struct input *input, struct key *key)
 }
 
 static enum input_status
+prompt_script_handler(struct input *input, struct key *key)
+{
+	switch (key_to_value(key)) {
+	case KEY_RETURN:
+	case KEY_ENTER:
+	case '\n':
+		return INPUT_STOP;
+
+	default:
+		return INPUT_OK;
+	}
+}
+
+static enum input_status
 prompt_yesno_handler(struct input *input, struct key *key)
 {
 	unsigned long c = key_to_unicode(key);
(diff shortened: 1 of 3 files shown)


src/prompt.c
```diff
@@ -1069,17 +1083,23 @@ exec_run_request(struct view *view, struct run_request *req)
...
+
 	if (cmd && *cmd && !argv_from_string(argv, &argc, cmd)) {
 		report("Too many arguments");
 		return REQ_NONE;
 	}
 
 	if (!cmd || !*cmd) {
-		report_clear();
+		if (!is_script_executing())
+			report_clear();
```
jonas: Is this conditional behavior needed?

jonas:
How much slower was it? Persistence in the user Tig history file is definitely a big.

src/prompt.c
rolandwalker: No, this part is not important.

rolandwalker:
The speed penalty depends on configuration and can be extreme:

```bash
$ touch ~/.inputrc
$ printf 'set blink-matching-paren on\n' >> ~/.inputrc
$ TIG_SCRIPT=<(printf ':echo%%(commit)\n:echo%%(commit)\n:echo%%(commit)\n:echo%%(commit)\n:echo%%(commit)\n:quit\n') /usr/bin/time ./src/tig
        2.70 real         0.09 user         0.05 sys
$ git checkout script-sidestep-readline && make > /dev/null
$ TIG_SCRIPT=<(printf ':echo%%(commit)\n:echo%%(commit)\n:echo%%(commit)\n:echo%%(commit)\n:echo%%(commit)\n:quit\n') /usr/bin/time ./src/tig
        0.17 real         0.08 user         0.04 sys
```

src/prompt.c
```diff
@@ -127,6 +127,20 @@ prompt_default_handler(struct input *input, struct key *key)
 }
 
 static enum input_status
+prompt_script_handler(struct input *input, struct key *key)
```
jonas: I am not sure if defining this is actually necessary. `read_prompt_incremental` allows to pass a `NULL` handler.

rolandwalker: There might be a smarter way to do it, but the `NULL` handler doesn't recognize newlines.

jonas:
Thanks.

PR #704, opened 2017-07-31
ensure ordered writes to trace file

rolandwalker:
This fixes a race condition which can be exercised readily on OS X by

``` bash
for i in $(seq 1 100); do make test/main/filter-args-test; done
```

It's of little importance to users but probably manifested as an occasional mystery glitch in Travis.


jonas:
Great, just saw the build fail because of this!

PR #705, opened 2017-07-31
ignore script lines starting with # (comment)

rolandwalker:
Did not change doc.  In the absence of a Scripts section, the user's assumption is likely that scripts are similar to `~/.tigrc` files.


jonas:
Good to see these changes to the script handling.

PR #707, opened 2017-08-01
tighten up assert_vars testing

rolandwalker:
`assert_vars` was a bit vulnerable to false positives, and unable to test on whitespace.

 * take second arg of `assert_equals` for whitespace strict/ignore. Notes are still on trailing args.
 * rewrite the `assert-vars` inline script: less likely to trivially exit, save empty strings as literal `""`
 * run at least one test after `assert-vars` to ensure that steps-script completes

The last is a bit of a hack, though it harms nothing.  Better would be for the `assert_vars` shell function to accept an argument for the number of variables tested (number of lines in the vars file).

### DIFF
--- a/test/tools/libtest.sh
+++ b/test/tools/libtest.sh
@@ -228,19 +228,29 @@ ORIG_IFS=
 assert_equals()
 {
 	file="$1"; shift
+	whitespace_arg="ignore";
+
+	if [ "$#" -ge 1 ]; then
+		whitespace_arg="${1:-}"
+		shift
+	fi
+	if [ "$whitespace_arg" = strict ]; then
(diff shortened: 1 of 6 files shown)


test/tools/libtest.sh
```diff
@@ -228,19 +228,29 @@ ORIG_IFS=
 assert_equals()
 {
 	file="$1"; shift
+	whitespace_arg="ignore";
```
jonas: This default value threw me off while reviewing. Might have been clearer if it defaulted to `-w` and the value parsing was inside the `if [ "$#" -gt 1 ]`.

jonas:
Good stuff.

PR #708, opened 2017-08-01
WIP user-programmable registers

rolandwalker:
RFC.

Summary
 * each register is named by a single ASCII letter
 * `:set-register r %(commit)`
 * `:echo %(register:r)`
 * `!=(r)echo capture first line of arbitrary shell command`

Example: arbitrary diffs
 * `:bind main P :set-register c %(commit)`
 * `:bind main W :exec !git diff --color-words %(register:c) %(commit)`
 * <kbd>P</kbd> to pick commit
 * move to new commit
 * <kbd>W</kbd> to view arbitrary word diff

Example: search diff content
```bash
cat > ~/.tigrc.deepsearch.script
:set-register q Apple_Terminal
:exec =(c)git log -i --format='%H' -G '%(register:q)'
:goto %(register:c)
:enter
^D
```
 * `:bind main <Esc>/ :script ~/.tigrc.deepsearch.script`

Example 2 would work better if
 * `%(prompt)` was available within scripts: script would begin with `:set-register q %(prompt/)`
 * ordinary `/` search could accept expansions: script would end with `:eval /%(register:q)`

Incidentally my real usecase is like Example 2 but with a full-text sqlite index of diff content.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -312,6 +312,7 @@ TIG_OBJS = \
 	src/grep.o \
 	src/ui.o \
 	src/apps.o \
+	src/registers.o \
 	$(GRAPH_OBJS) \
 	$(COMPAT_OBJS)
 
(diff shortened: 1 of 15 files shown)


jonas:
Very interesting. I will probably wait with merging such a big change until I manage to release all the other good stuff from you.

PR #710, opened 2017-08-01
Disable leak detection for AddressSanitizer

jonas:


rolandwalker:
I don't understand the address sanitizer enough to know how I broke it, but the gcc failure here at the moment looks like the intermittent one fixed by #704 .

jonas:
I think it broke because Travis has upgraded the Linux builds to Trusty which has a newer version of clang.

jonas:
I pushed a fix to `master` based on the solution that Travis just validated.

PR #720, opened 2017-08-04
add --assert-equals=<filename>=<content> to test_case

rolandwalker:


rolandwalker:
Or should it be --assert-equals=<filename>=<content>?

PR #721, opened 2017-08-10
 Add the packaging metadata to build the tig snap

elopio:
This package will let you publish the latest tig in the Ubuntu store, and from there reach many users on all the supported Ubuntu versions, and more Linux distributions in progress. You just have to go to https://build.snapcraft.io and enable the automated continuous delivery.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,19 @@
+name: tig
+version: git
+summary: Text-mode interface for git
+description: |
+  Tig is an ncurses-based text-mode interface for git. It functions mainly as a
+  Git repository browser, but can also assist in staging changes for commit at
+  chunk level and act as a pager for output from various Git commands.
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: classic
+
+apps:
+  tig:
+    command: tig
+
+parts:
+  tig:
+    source: .
+    plugin: autotools
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -11,7 +11,7 @@ confinement: classic
 
 apps:
   tig:
-    command: tig
+    command: bin/tig
 
 parts:
   tig:


elopio:
Ping, @jonas any chance you can review this?

elopio:
sorry, wrong button :blush:

simos:
1. You use `classic` confinement. Perhaps it's better to use `strict` confinement?
It would then work only on repositories on  ~/ but not be able to read ~/.gitconfig (for now).

2. `tig` requires `git` to work. It does not require the full `git` but just a small part of it.
`git` could be compiled and provided in the `tig` package.

jonas:
@simos I'm new to snap. I'll go ahead and merge this as is, but please update if it makes sense to tweak what tig can access.

simos:
@jonas I just checked that `tig` is available. Go ahead and register it.

You run `snapcraft register tig`. You will then be asked:

    Would you say that MOST users will expect 'tig' to come from
    you, and be the software you intend to publish there? [y/N]:

and you type **y**.

simos:
@elopio Here is my snapcraft.yaml for `tig`. Could you please comment on it?

```
name: tig
version: '2.3.3'
summary: Text-mode interface for git 
description: |
  Tig is an ncurses-based text-mode interface for git. It functions
  mainly as a Git repository browser, but can also assist 
  in staging changes for commit at chunk level and act as 
  a pager for output from various Git commands.

grade: stable
confinement: strict

apps:
  tig:
    command: tig
    plugs:
      - home

parts:
  tig:
    source: https://github.com/jonas/tig.git
    source-tag: tig-2.3.3
    source-depth: 1
    plugin: autotools
    build-packages:
      - libncursesw5-dev
  git:
    source: https://github.com/git/git.git
    source-tag: v2.17.0
    source-depth: 1
    plugin: autotools
    build-packages:
      - libz-dev
      - gettext

```

Specifically,

1. `tig` uses `git`, so this snap grabs, compiles a recent `git` and puts it in the snap. 
`tig` does not require much of the utilities in `git`, so most could be removed to cut down the size of the snap. @jonas, does `tig` use only the `git` binary?

1. This snapcraft.yaml does not use `classic` confinement, therefore it can get added to the Snap Store without having to go through asking special permission. @jonas can add it straight away.
The downside is that the repositories must be located in the home directory of the user (due to the security confinement issues). 
@jonas, if you are not happy with this and want the `tig` snap to work anywhere on the filesystem, then go for the `classic` confinement setting.

3. Also, in the `strict` confinement, `tig` will not be able to read ~/.gitconfig. As far as I know, the ability to read *dot* configuration files has not been sorted yet.

jonas:
I requested `tig` as the snap name.

 1. At runtime `tig` requires git, libncurses, libiconv (if not provided by libc) and libreadline (optional).
 2. I think it makes sense to allow repositories located outside of `$HOME`.
 3. For my own use not having access to `~/.gitconfig` makes Tig less useful, since you cannot launch for example `git commit --amend`.

simos:
1. A system that supports snap packages, has a common snap package with shared libraries called `core`. That is, any snap package can make use of libraries in `core` and not need to include them themselves. You can see what's in `core` by searching under `/snap/core/current`. And `libncursesw` (UTF-8 support) can be found in there, `/snap/core/current/lib/x86_64-linux-gnu/libncursesw.so.5`.
The same goes with `libiconv`, it is found in `core`'s libc. Also, libreadline is at `/snap/core/current/lib/x86_64-linux-gnu/libreadline.so.6`.

The issue is with `git`, which needs to be added by `tig`. 

2. In that case, it's necessary to use `confinement: classic`. For this, you apply to enable this type of confinement. 

Note that $HOME access would include access to `gvfs` mounted filesystems and you could also add access to `FUSE`-mounted filesystems. But other paths would not be accessible, therefore `classic` would be required.

3. There is new support for dot file access (https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1607067), but if you get `classic` support, you would be fine without specifying dot files.

PR #723, opened 2017-08-12
quieten "unary argument" warning during configure

rolandwalker:
shell variable `$bash_cv_termcap_lib` can be empty


jonas:
Thanks

Issue #301, opened 2014-06-06
Feature to auto refresh tig status if any file changes


jaseemabid:
Can't this be done with inotify?

rolandwalker:
A portable external tool is [entr](http://entrproject.org/).

A first approximation of entr monitoring a repo is

```bash
git ls-files "$(git rev-parse --show-toplevel)" | entr -pd printf '%s\n' /_
```

with the only wrinkle being that entr has two modes of action: upon modification to the listed files, it executes the given command (`printf`); upon other modifications to the directories containing the listed files, it exits.

An interesting thought is

```bash
while true; do git ls-files "$(git rev-parse --show-toplevel)" | entr -pd kill -USR1 "$TIG_PID"; kill -USR1 "$TIG_PID"; done
```

msudgh:
Did this feature implement? what's the current status? I'd like to have a PR about this feature.

rolandwalker:
Some time ago I sketched out an implementation indea in #734.  It seems like the momentum got lost there.  There is an unexplained test failure.

However, I have been using that branch in my daily work for a long time without any problems.

koutcher:
After #917 the legacy refresh modes should now work as expected. Is this new one still necessary ?

msudgh:
@koutcher Currently I'm using tig `2.4.1` and I don't see any periodic refresh which still needs to press `Shift+R` to refresh the status layout.

koutcher:
It will be in the next version, for now you'd need to compile the master branch to use it.

msudgh:
Thanks @koutcher

gdgp:
Built from master, for some reason `set refresh-mode = auto` was not working for me. However the following is working:
```
set refresh-mode = periodic
set refresh-interval = 1
```

koutcher:
What you should expect when `set refresh-mode = auto` is that after opening the status view and staging a file, the change is also visible in main view, nothing more. If you want tig to actively look for changes in the directory then you got the right choice with `periodic`.

gdgp:
@koutcher I see, thank you for clearing that up. Then I think I got mislead from reading in [tigrc.5@master](https://github.com/jonas/tig/blob/master/doc/tigrc.5.adoc) the following:
"refresh-mode (mixed) [manual|auto|after-command|periodic|<bool>]
[...]

PR #725, opened 2017-08-14
save and restore TTY attributes

rolandwalker:
main program:

 * `opt_tty` becomes a struct
 * `tcgetattr()` at init time
 * `tcsetattr()` in `done_display()`
 * specify `TCSAFLUSH` action in `done_display()` so that tig can't leave keystrokes behind

`open_external_viewer()`, after running external command:

 * restore init-time tty attributes before duplicating ncurses setup steps
 * use `TCSAFLUSH` action when restoring tty attributes to block keystrokes leaking in from external command phase - probably more robust than `fseek()` on the file pointer
 * but keep `fseek … SEEK_END`, which does no harm, make it unconditional

edit: also set tty attributes and flush before running external command

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -30,7 +30,12 @@ static WINDOW *display_win[2];
 static WINDOW *display_title[2];
 static WINDOW *display_sep;
 
-static FILE *opt_tty;
+struct display_tty {
+	FILE *file;
+	int fd;
+	struct termios *attr;
+};
+static struct display_tty opt_tty = { NULL, -1, NULL };
 
 static struct io script_io = { -1 };
 
@@ -83,8 +88,8 @@ open_external_viewer(const char *argv[], const char *dir, bool silent, bool conf
 			if (!ok && *notice)
 				fprintf(stderr, "%s", notice);
 			fprintf(stderr, "Press Enter to continue");
-			getc(opt_tty);
-			fseek(opt_tty, 0, SEEK_END);
+			getc(opt_tty.file);
+			fseek(opt_tty.file, 0, SEEK_END);
 		}
 		set_terminal_modes();
 	}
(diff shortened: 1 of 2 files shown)


src/display.c
```diff
@@ -566,15 +579,32 @@ set_terminal_modes(void)
...
+}
+
 void
 init_display(void)
 {
 	bool no_display = !!getenv("TIG_NO_DISPLAY");
 	const char *term;
 	int x, y;
 
+	init_tty();
```
jonas: This will open `/dev/tty` for tests. I think it would be better to skip this if `no_display` is true.

rolandwalker: Intentionally so — from the point of view of the OS, a TTY is much more than a display.

The TTY is also the medium for the line discipline, in rough terms the interactive input.

A TTY is also inseparable from process-group leadership via `tcsetpgrp()`, which among other things determines the disposition of signals, which is relevant under the test harness.

Also, regardless of whether a display is visible, tig is still linked with ncurses, and ncurses maintains a thoroughly TTY-oriented understanding of the world in which it runs.   The goal I was driving for here is [for ncurses to keep an input-only TTY under `no_display`](https://github.com/jonas/tig/pull/736/commits/55e84f0488c4b4510ec9ebc6dd923205cafecced).  Tig still does not *display* anything, but the input "side" of ncurses remains connected to a proper [`ioctl`able TTY](https://github.com/jonas/tig/pull/736/commits/e2a2910cef2bf00d8a99789c8ab4bc2a117fc960#diff-2a68c783d9ab596cb14680c33e4cf5acR421).

The separate matter of process-group leadership becomes relevant when there are long-lived child processes as proposed in #301/#734.  The simplest and most robust way to handle child cleanup is to
 * be process-group leader
 * [`SIGHUP` your own process group at exit](https://github.com/jonas/tig/pull/734/files#diff-28d128864ec9d27d6bdea2ab4c987955R717).

In the most ordinary case, the interactive shell makes tig the process-group leader transparently, and we don't have to think about it.  However, that magic does not happen in non-interactive cases, such as when driven by `tig-pick` or the test harness (where `no_display` happens to be set).  So it is a good idea, when you know you may be invoked from a script, to grab a TTY as a [step toward becoming process-group leader](https://github.com/jonas/tig/pull/736/commits/e822119435b2b7315ec640e2fd9d13ff51c5c969).

Congrats on the release!

jonas:
Thanks a lot for writing down the detailed motivation.

PR #727, opened 2017-08-15
save/restore TTY attributes in test suite

rolandwalker:
defending against a tig crash

### DIFF
--- a/test/tools/libtest.sh
+++ b/test/tools/libtest.sh
@@ -28,6 +28,7 @@ output_dir="$base_dir/tmp/$prefix_dir/$test"
 tmp_dir="$base_dir/tmp"
 output_dir="$tmp_dir/$prefix_dir/$test"
 work_dir="work dir"
+tty_attrs="$(stty -g </dev/tty)"
 
 # The locale must specify UTF-8 for Ncurses to output correctly. Since C.UTF-8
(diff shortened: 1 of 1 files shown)


jonas:
What's the motivation for this? Tests are not supposed to touch the TTY.

rolandwalker:
I believe the motivation is fairly covered in https://github.com/jonas/tig/pull/725#discussion_r142008745, but even without #725, there were some cases where a tig crash during tests left the TTY in a wrong state.

Ncurses may do some unexpected things to resolve conflicts: *eg* tig gives ncurses `/dev/null` for the third argument to `newterm()`, but then tig calls `raw()` and so on.  And `raw()` must call `ioctl()`, which is only possible on an fd which refers to a device or special file (tty).  `raw()` might be returning `ENOTTY` and we fail to check the return value.

But in at least some cases it seems that ncurses decides to muddle through with some undefined behavior.  For example it switches to STDIN for `wgetch()` after being told that `infd` referred to `/dev/null`.

That is from memory, and I don't have the code in front of me, so some details may be wrong.

Feel free to decline the PR — it is just a bit of defensive driving.  The explanation is for your interest.

jonas:
Ah, OK. Thanks for the explanation. Makes sense.

PR #729, opened 2017-08-15
consistency in use of IFS

rolandwalker:
* remove bash-dependent setting
 * always use space+tab when wordsplitting is wanted
 * set IFS for `git diff` where worksplitting is wanted (bug)

As a matter of opinion, I think it is better not to manipulate IFS at all but to follow shell quoting rules.


jonas:
Thanks

Issue #490, opened 2016-04-03
graph_remove_collapsed_columns heap-buffer-overflow


comicfans:
line 405
for (i = row->size - 1; i > 0; i--) {
             ....

                if (strcmp(row->columns[i - 1].id, graph->prev_row.columns[i - 1].id) != 0 || graph->prev_row.columns[i - 1].symbol.shift_left)
                        row->columns[i] = row->columns[i + 1];   -----------------> i+1 may overflow row->size
                }

ASAN report this when tig running in linux kernel directory

==14616==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x621000736300 at pc 0x46c729 bp 0x7fffffffd810 sp 0x7fffffffd800
READ of size 48 at 0x621000736300 thread T0
    #0 0x46c728 in graph_remove_collapsed_columns src/graph-v2.c:434
    #1 0x46c9e0 in graph_generate_next_row src/graph-v2.c:457
    #2 0x46ec2a in graph_render_parents src/graph-v2.c:719
    #3 0x462fea in main_read src/main.c:437
    #4 0x43a0a0 in update_view src/view.c:550
    #5 0x4350f3 in get_input src/display.c:581
    #6 0x42bf78 in prompt_input src/prompt.c:49
    #7 0x42cabc in read_prompt_incremental src/prompt.c:178
    #8 0x4068a1 in read_key_combo src/tig.c:644
    #9 0x406dce in main src/tig.c:707
    #10 0x7ffff66c659f in __libc_start_main (/lib64/libc.so.6+0x2059f)
    #11 0x403978 in _start (/root/.local/bin/tig+0x403978)

0x621000736300 is located 0 bytes to the right of 4608-byte region [0x621000735100,0x621000736300)
allocated by thread T0 here:
    #0 0x7ffff6f56ab6 in __interceptor_realloc (/usr/lib/gcc/x86_64-pc-linux-gnu/4.9.3/libasan.so.1+0x57ab6)
    #1 0x40b8e9 in chunk_allocator src/util.c:370
    #2 0x46a5db in realloc_graph_columns src/graph-v2.c:79
    #3 0x46b372 in graph_insert_column src/graph-v2.c:274
    #4 0x46b916 in graph_expand src/graph-v2.c:318
    #5 0x46ec0d in graph_render_parents src/graph-v2.c:716
    #6 0x462fea in main_read src/main.c:437
    #7 0x43a0a0 in update_view src/view.c:550
[...]

PR #730, opened 2017-08-16
stub docs for testing API

rolandwalker:
`test_setup()` doesn't look like it is meant to be exposed, but is called by `gh-490-heap-buffer-overflow-test`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ RPM_RELEASE = $(word 2,$(RPM_VERLIST))$(if $(WTDIRTY),.dirty)
 DFLAGS	= -g -DDEBUG -Werror -O0
 EXE	= src/tig
 TOOLS	= test/tools/test-graph tools/doc-gen
-TXTDOC	= doc/tig.1.adoc doc/tigrc.5.adoc doc/manual.adoc NEWS.adoc README.adoc INSTALL.adoc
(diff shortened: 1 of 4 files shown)


jonas:
Very good idea, I'll try to see if I can fill it out. `test_setup` is there to exit quickly in the test, we might be able to remove if we add an `exit 0` call in `test_skip`.

jonas:
To make this manageable I suggest to embed the documentation into `libtest.sh` and `libgit.sh` using docstrings and have a small too that extracts them to `test/api.md`.

Something like
```sh
## Run tig in test mode
#
# Setup a controlled environment and report the test result.
#
# Input to be procssed via stdin is passed and stderr is captured and
# can be used for later assertions.
#
# ```sh
# test_tig show 1a2b3c4d5e6f
# ```
test_tig()
{
...
```

Which would become:

```markdown
## test_tig: Run tig in test mode

Setup a controlled environment and report the test result.

Input to be procssed via stdin is passed and stderr is captured and
can be used for later assertions.

```sh
test_tig show 1a2b3c4d5e6f
`` `
```

rolandwalker:
Yes good idea.

rolandwalker:
reimplemented as `make test/API.adoc`

jonas:
Thanks

PR #731, opened 2017-08-16
bug: branch on content in expected file

rolandwalker:
If the user didn't specify content for STDERR, that is the same as specifying it must be empty


jonas:
Good catch

PR #735, opened 2017-08-18
convert test timeout watchers to polls

rolandwalker:
so that they don't linger after running tests

### DIFF
--- a/test/tools/libtest.sh
+++ b/test/tools/libtest.sh
@@ -487,7 +487,17 @@ install_pid_timeout() {
 	test "$pid" -gt 0 || return
 	test "$pid" != "$$" || return
 	trap '' "$signal"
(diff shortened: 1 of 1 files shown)


jonas:
Can you enlighten me again why you wanted the test timeouts in?

rolandwalker:
Primarily, to add tests that work by simulating keystrokes to the tty.

Just stuffing the keystrokes is simple tech, and I have it working in a private branch.

But compared to scripts it is easier for an arbitrary series of keystrokes to leave tig in a state where naively appending `:quit\n` would not successfully exit.  Timeouts are a backstop.

This is also a motivator for #725 and #727 .

jonas:
Thanks for the explanation. Great to have the lingering fixed.

PR #736, opened 2017-08-23
simulated keystrokes in test suite

rolandwalker:
Summary

```bash
keystrokes                 ':none'
keystrokes -append -keysym 'Enter'
keystrokes -append         ':quit'
keystrokes -append -keysym 'Enter'
# or keystrokes ':none\r:quit\r'

test_tig

assert…
```

Further documented in ~test/README.adoc~ `test/API.adoc`.

~This is in reasonably functional form but is a WIP in part because it will merge conflict with several other outstanding PRs.~  This is in functional form and merge conflicts have been resolved.

Another interesting approach to simulating input is driving tig inside of tmux as in

 * https://github.com/zsh-users/zsh-autosuggestions/blob/master/spec/terminal_session.rb

this has the advantage that the screen-capture action can be done by tmux.  Though leveraging that would likely mean changing all existing tests.

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -611,7 +611,6 @@ init_display(void)
 {
 	bool no_display = !!getenv("TIG_NO_DISPLAY");
 	const char *term;
-	int x, y;
 
 	if (!opt_tty.file)
 		die("Can't initialize display without tty");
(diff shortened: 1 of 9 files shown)


rolandwalker:
Now that tig is the process-group leader this PR is more focused and can be evaluated on its own.

I suspect that you won't like the added complexity and the python requirement (though keystroke test are skipped without python.)

But it was fun to do in any case.  The motivation was to provide test coverage to some hidden corners that libtest scripts could not reach.

PR #737, opened 2017-08-25
add ability to select tests/cases via TEST_OPTS='filter=<glob>:<glob>'

rolandwalker:
and give feedback as test cases are run:

```
TEST  test/tigrc/width-test
CASE  test/tigrc/width-test:main-widths-default
CASE  test/tigrc/width-test:main-widths-1
CASE  test/tigrc/width-test:main-widths-2
…
```


jonas:
Sweet ❤️

Issue #463, opened 2015-12-09
"Press Enter to continue" after a commit message


solarsailer:
Hello,

I don't know if it's tig's problem or Vim or git, but when I commit my changes in the status view, tig asks me to press enter to go back in the status view after writing the commit message:

![image](https://cloud.githubusercontent.com/assets/478621/11693533/b359282c-9ea5-11e5-931e-78954e85e6a5.png)

Is there a way to disable this behavior? I want to re-open the status view ASAP without my intervention. I can't find a way to specify this (either in my `.tigrc` or somewhere else).

Thanks.

jonas:
Right now there's no way to disable it. We can add a new open flag for user-defined commands. Maybe `>` as the opposite of `<` which currently means "exit Tig after command". This would allow you to write:

    bind generic C >git commit

solarsailer:
What would it mean exactly ?

```
>     Re-open Tig instantly in the last displayed view after executing the command.
```

Something like that? Would be perfect. :)

DiegoPomares:
I use difftool (vimdiff) to check diffs and the message annoyed me. Here's the pull request that implements >

https://github.com/jonas/tig/pull/738

PR #738, opened 2017-08-25
Issue #463 Re-open Tig instantly in the last displayed view after exe…

DiegoPomares:
Implemented the > flag, I called it 'quick'  internally:

- I couldn't think of any tests to add :/
- The message (press Enter to continue) is still shown if ok in 'ok = io_run_fg(...)' is false (line 86@src/display.c), but maybe it is better to hide it as well even if ok is false. I'll leave that to you.
- Perhaps it would be better to use the quick mode for the merge tool as well (line 679@src/status.c), but I left it as it was. I'll also leave that to you.


Alok:
Any chance of this getting merged soon/

aesnyder:
Curious if this will be merged as well.

DiegoPomares:
Dunno @aesnyder @alok :man_shrugging:
BTW guys since you seem interested, what do you think about my original remarks about not showing the _press Enter to continue_ message when the command exits with a non 0 status, and for the merger tool?

alok:
@DiegoPomares I'm not sure about the nonzero exit code case, but for merge I think it's fine.

DiegoPomares:
@alok That was my idea initially, thanks for the feedback :smile:

alok:
@jonas any change of this getting added?

src/status.c
```diff
@@ -676,7 +676,7 @@ open_mergetool(const char *file)
 {
 	const char *mergetool_argv[] = { "git", "mergetool", file, NULL };
 
-	open_external_viewer(mergetool_argv, repo.cdup, false, true, false, true, "");
+	open_external_viewer(mergetool_argv, repo.cdup, false, true, false, true, true, "");
```
jonas: Not sure what the motivation is for using the quick mode here. I'll change it to the original behavior.

jonas:
Thanks a lot @DiegoPomares ! 🎉

Issue #663, opened 2017-07-13
Segfaulting on master


edi9999:
Hello,

I was using tig-2.2.1 before, and it never segfaulted.

Now on almost master : a628e9d1680d8e9a1ffeba60bd07db24f993bbf0, I see segfaults sometimes when I start tig. It happens immediately on startup, I don't even have the time to see the main view.

I cannot reproduce the bug however, it seems to happen quite randomly.

Do you know how I could give more information ? I have tried with gdb, but it never segfaults then.

jonas:
Thank you for the report.

If you have time and it "semi-reliably" segfaults it would be very helpful if you could do a `git bisect`. Else I can try to add in some code to generate a stack trace to the terminal.

edi9999:
It doesn't segfault with 2.2.2 (I'm on it since these 3 days)

edi9999:
I have switched to master again because of the new `+` feature which I really appreciate. However it is segfaulting quite often.

Can you add the code to generate a stack trace to the terminal please ?

jonas:
@edi9999 I just pushed a commit 8c0a7c0acc13c019517dc1d9f3c9bcb0cd3a08e3 which allows you to have stacktraces. However, you need to compile Tig with `make all-debug`. Hope that will help us get the info required to fix this issue.

edi9999:
There was a crash after I installed it, however it doesn't show any useful info I think : 

```
Tig crashed!

Please report this issue along with all info printed below to

  https://github.com/jonas/tig/issues/new

Tig version: 2.2.2-160-g8c0a7c0-dirty

Aborted (core dumped)
```

jonas:
Ah, can you try to build using this command then: `make all-debug CPPFLAGS=-DHAVE_EXECINFO_H`

jonas:
Else start Tig via the debugger:

    > gdb ./src/tig
    gdb> run

When it segfaults run `bt` in the gdb shell to get the backtrace.

edi9999:
It seems to be some race condition of some sort, I was never able to reproduce it with gdb, also, it seems to happen 5% of the time, so not so easy to reproduce.

It just did it again, with a stacktrace

```
Tig crashed!

Please report this issue along with all info printed below to

  https://github.com/jonas/tig/issues/new

Tig version: 2.2.2-160-g8c0a7c0-dirty

tig[0x40452d]
/lib64/libc.so.6(+0x366f0)[0x7f9fc3b5e6f0]
tig[0x434993]
[...]

PR #745, opened 2017-09-15
Fix segfault with htab_empty

edi9999:
Refs: #663


edi9999:
The segfault happened 1 time out of 5 on my machine, at the start of `tig`

PR #750, opened 2017-10-03
parallelize build using travis

stephengroat:
can also use build stages if needed (only start building autoconf after config has succeeded)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,12 +23,13 @@ script: tools/travis.sh
 
 matrix:
   include:
-  - os: osx
+  - if: branch = master
+    os: osx
     env:
     compiler: clang
     script: brew install --HEAD tig
-  - os: osx
+  - if: branch = master
+    os: osx
     env:
     compiler: gcc
     script: brew install --HEAD tig
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 language: c
 
-compiler:
-  - clang
-  - gcc
-
 sudo: false
 
 addons:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -13,25 +9,61 @@ addons:
...
-  - ./configure --prefix=/tmp/conf-prefix
-  - make V=1 TEST_SHELL=bash all test install install-doc
-  - make uninstall
-  - test ! -d /tmp/conf-prefix
-  - make clean
+matrix:
+  include:
+  - compiler: clang
+    script:
+    - # ============ Build from config ===============
```
jonas: Would you mind to also move the script code into `tools/travis.sh` which based on the argument passed will run one of the builds?

.travis.yml
```diff
@@ -13,25 +9,61 @@ addons:
...
+    - # ============ Build using autoconf ============
+    - make dist
+    - ./configure --prefix=/tmp/conf-prefix
+    - make V=1 TEST_SHELL=bash all test install install-doc
+    - make uninstall
+    - test ! -d /tmp/conf-prefix
+    - make clean
+  - os: osx
+    compiler: clang
+    script: brew install --HEAD tig
```
jonas: I don't think it makes sense to build on osx and less so using brew. In my experience, osx builds are slow, and we don't check compatibility for other package managers.

stephengroat: `osx` builds are definitely slow, but i think that you have a significant user base in osx (https://bintray.com/homebrew/bottles/tig#statistics). when `brew` bottles (the binary releases) are built, no tests are run. just a way to get some sanity.

i'll remove if you think it's best (you can also make them `allow_failures`), but i think it gets more coverage for free from probably the most popular `osx` package manager

jonas: Wow, cool. Didn’t know about those statistics. OK let’s keep one osx build then. However, then I wonder if `—head` will build the current branch. If this is not the case then the osx build should only trigger when the master branch changes. WDYT?

stephengroat: your right, `--HEAD` builds `master` only. i'll throw that in and get the other changes done

stephengroat:
think i got what you wanted in, let me know if any other changes are necessary

jonas:
Thanks, looking forward to trying this out.

PR #758, opened 2017-11-04
Handle \n like \r

chrisbra:
Currently, when \n is received, tig will report
"Unknown key, press h for help"

This is a workaround for vim/vim#1998 but nevertheless it might make
sense to handle \n synonymous to \r. So this patch makes handling \n
like \r.

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -813,6 +813,9 @@ get_input(int prompt_position, struct key *key)
 			input_mode = false;
 			if (key_value == erasechar())
 				key_value = KEY_BACKSPACE;
+			/* Handle \n just like \r */
+			else if (key_value == '\n')
+				key_value = KEY_RETURN;
 
 			/*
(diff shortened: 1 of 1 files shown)


jonas:
Thanks @chrisbra

0ax1:
This seems to break the distinction between `<c-j>` and `<enter>`.  In effect, `<c-j>` key bindings stopped working with the introduction of this commit (found via `git bisect`).

jonas:
@0ax1 You are right. I hadn't thought about this and we don't have a test case for C-j it seems. Do you want to make a PR that reverts this change?

My understanding is that Him fixed this so this workaround is not necessary.

0ax1:
@jonas Sure, I'll open a PR.

chrisbra:
Upps, sorry for causing this.

0ax1:
@chrisbra No prob. :), I just pushed a fixup PR already.

PR #759, opened 2017-11-05
adjust case of reldate letters acc. to ISO-8601

plgruener:
This commit fixes the wrong use of 'm' for month and 'M' for minute:
In ISO-8601, uppercase letters YYYY-MM-DD are used to denote year,month,week,day;
whereas lowercase letters hh:mm:ss are used for hour,minute,second.


jonas:
Sorry, it's been a while. The `test/main/date-test` files needs to be updated. Do you want to take a stab at that?

plgruener:
Yeah, I'll try to do it during the holidays.
But please be aware, as written in #760 , that this change will probably break user's custom tigrcs which do a fancy date-setting.

jonas:
Custom date formats must use the format expected by `strftime(3)`.
This PR only changes the way relative dates are displayed so no user configuration will be affected.

jonas:
When you run the `date-test`:

```
> make test/main/date-test
      TEST  test/main/date-test
```

You can see the output of the actual screen with the relative date in the file

    test/tmp/main/date-test/relative-compact-date.screen

jonas:
Thanks, I fixed the test and merged.

PR #778, opened 2018-01-31
Fix the resource leak

hwangcc23:
Fix the file descriptor leak.


hwangcc23:
Have some other mirror-fix patches.
Will send all patches in another PR later.
Close this first.

jonas:
Thanks, sorry for taking so long to look at them.

Issue #781, opened 2018-02-12
"Edit" function in tree and diff view opens the wrong path under certain conditions in submodules


h3xx:
## Part One

When the git repo you're viewing is cloned as a submodule, `e` opens the wrong file path for editing, **but only if your working directory isn't the top directory of the submodule** if that makes any sense.

Steps to reproduce:

1. Run these shell lines:

    $ mkdir temp ; cd temp
    $ git init
    $ git submodule add https://github.com/jonas/tig.git
    $ cd tig/src
    $ tig

2. Then in the resulting tig window, press enter on a commit, then use j/k to highlight a diff-stat line and press e. It will give the error "Failed to open file: ..."

Expected behavior: Open the file in $EDITOR from the highlighted line in the diff view.

## Part Two

Likewise, pressing `e` on a file when in tree view results in the editor opening a non-existent file, **but only sometimes**. It seems to have to do with it being both a submodule, and trying to open a file using tree view **two levels above the directory where you started tig.**

Steps to reproduce, part 2:

    $ mkdir temp ; cd temp
    $ git init
    $ git submodule add https://github.com/jonas/tig.git
    $ cd tig/tests/graph
    $ tig

2. Press `t` to enter tree view, then press enter on `..` twice, so you're at the top level for the submodule.
3. Highlight a file in that directory like `INSTALL.adoc` and press `e`.

I find it will open `temp/INSTALL.adoc` which doesn't exist, instead of `temp/tig/INSTALL.adoc`.

Tested under 58aa1d4f215855be78671ffc41d4259b77839b9b running on Slackware Linux x86_64 with git-2.16.1.

h3xx:
See f99d2afbfaafcefcbc9ad41b3d7d6919854e7a01 for a test that proves it's an issue.

h3xx:
Confirmed fixed. Thanks!

jonas:
Thanks for testing.

PR #783, opened 2018-02-13
Fix opening editor from diff-stat line

h3xx:
Partial fix for #781

This still doesn't fix the underlying issue where the working directory gets set incorrectly when in a submodule.


jonas:
Thanks a lot. It would be nice to have a test for this since sub modules are not “well” supported. I will try to find some time.

h3xx:
See f99d2afbfaafcefcbc9ad41b3d7d6919854e7a01 for a related unit test.

h3xx:
Confirmed fixed. Thanks!

PR #784, opened 2018-02-18
Fix some problems

silvio:
Hi,

have tested some linters with your project and this two patches are the outcome of this tests.

### DIFF
--- a/src/prompt.c
+++ b/src/prompt.c
@@ -1047,7 +1047,7 @@ run_prompt_command(struct view *view, const char *argv[])
 			prompt_update_display(flags);
 
 		} else {
-			request = view_can_refresh(view) ? REQ_REFRESH : REQ_SCREEN_REDRAW;
+			view_can_refresh(view) ? REQ_REFRESH : REQ_SCREEN_REDRAW;
 			if (!strcmp(cmd, "color"))
 				init_colors();
 			resize_display();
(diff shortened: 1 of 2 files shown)


src/prompt.c
```diff
@@ -1047,7 +1047,7 @@ run_prompt_command(struct view *view, const char *argv[])
 			prompt_update_display(flags);
 
 		} else {
-			request = view_can_refresh(view) ? REQ_REFRESH : REQ_SCREEN_REDRAW;
+			view_can_refresh(view) ? REQ_REFRESH : REQ_SCREEN_REDRAW;
```
jonas: This line can be completely removed.

src/view.c
```diff
@@ -1667,9 +1668,10 @@ append_line_format(struct view *view, struct line *line, const char *fmt, ...)
 	if (fmtlen <= 0)
 		return false;
 
-	box = realloc(box, box_sizeof(box, 0, fmtlen));
-	if (!box)
+	temp_box = realloc(box, box_sizeof(box, 0, fmtlen));
```
jonas: ‘box’ capture the value in ‘line->data’, so this assignment is not supposed to lose anything.

jonas:
Thanks, for the contribution. I'm closing this as it seems to have gone stale.

PR #788, opened 2018-03-06
Fix file argument

koutcher:
file -I works with macOS but not with Linux (should be file -i). Rather use the long option --mime which is valid for both.


jonas:
Thanks for the fix. Will try to get a new version out soon.

PR #789, opened 2018-03-10
display.c: fix build with netbsd-curses

rofl0r:
refresh is implement as a macro in netbsd curses,
so using it as a variable name clashes.


jonas:
Thanks for fixing this.

PR #790, opened 2018-03-10
display.h: fix build against netbsd-curses

rofl0r:
`newscr` is a ncurses-ism, and doesn't exist in netbsd-curses.

according to `man 3x curs_variables`:

> The New Screen
> This implementation of curses uses a special window newscr to hold  up-
> dates to the terminal screen before applying them to curscr.

so it seems using curscr instead should be fine.

### DIFF
--- a/include/tig/display.h
+++ b/include/tig/display.h
@@ -58,8 +58,8 @@ void enable_mouse(bool enable);
 
 enum status_code open_script(const char *path);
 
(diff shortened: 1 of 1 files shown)


jonas:
Merging so it is easier to test it out. Again thanks a lot!

rofl0r:
thanks. please let me know if this one causes issues.

PR #791, opened 2018-03-11
New boolean option "send-child-enter"

andebjor:
Can be set to "no" to disable sending enter key presses to child view.
In practice only relevant if "focus-child" is false.

Default is "yes", which is identical to previous behavior.


jonas:
You are not the first to raise this issue, much appreciated that you took the time to make a patch.

PR #792, opened 2018-03-12
Update make config for cygwin/ncurses

andebjor:
Cygwin has apparently updated the `ncursesw` package to version 6.
Update make config to reflect that.


jonas:
Thanks a lot.

PR #797, opened 2018-03-17
Rework #790 to define newscr in tig.h if undefined

jonas:


jonas:
@rofl0r Would it be possible for you to test this? I can merge if that makes it easier.

rofl0r:
that looks as if it would work, but i can test it later. why do you think this change is needed ? was there a regression ?

rofl0r:
```
src/tig.c: In function 'parse_options':
src/tig.c:517:21: error: #if with no expression
 #if NCURSES_WIDECHAR
                     ^
```
i suppose this should be `#if NCURSES_WIDECHAR+0`

jonas:
Thanks for testing this @rofl0r

PR #799, opened 2018-03-19
Warning fixes

proski:
I tried compiling tig with most warnings enabled using the current gcc snapshot that would become gcc 8. I started with `-Wall -Wextra` and disabled warnings that cannot be easily fixed (`-Wno-unused-parameter -Wno-sign-compare -Wno-missing-field-initializers`)

I was able to fix all warnings, but some fixes need more work, so I'm only submitting the simple and good-looking fixes.


jonas:
Thanks for fixing the warnings.

Issue #800, opened 2018-03-19
Valgrind and ASAN detect invalid read 8 bytes before allocated region at ui.c:143


proski:
Here's how to reproduce the issue.

Compile tig (current master from git).
Run `valgrind --log-file=valgrind.log src/tig`
t (tree view)
select `tigrc`
b (blame)
f (show file)
f (find file)
fff (entering the file name)
escape
q (repeat until tig exits)

From valgrind log:
```
==22191== Invalid read of size 8
==22191==    at 0x4231CE: file_finder_draw (ui.c:143)
==22191==    by 0x42354D: file_finder_input_handler (ui.c:271)
==22191==    by 0x410B43: prompt_input (prompt.c:67)
==22191==    by 0x4237A3: open_file_finder (ui.c:302)
==22191==    by 0x41D21D: open_blob_view (blob.c:40)
==22191==    by 0x4039B3: view_driver (tig.c:221)
==22191==    by 0x4039B3: main (tig.c:799)
==22191==  Address 0xcaa2ae8 is 8 bytes before a block of size 2,048 alloc'd
==22191==    at 0x4C2CAAC: malloc (vg_replace_malloc.c:298)
==22191==    by 0x4C2EC5F: realloc (vg_replace_malloc.c:785)
==22191==    by 0x405A9D: chunk_allocator (util.c:390)
==22191==    by 0x4236FC: realloc_file_array (ui.c:27)
==22191==    by 0x4236FC: file_finder_read (ui.c:76)
==22191==    by 0x4236FC: open_file_finder (ui.c:286)
==22191==    by 0x41D21D: open_blob_view (blob.c:40)
==22191==    by 0x4039B3: view_driver (tig.c:221)
==22191==    by 0x4039B3: main (tig.c:799)
```

Now let's try ASAN.
```
make CFLAGS="-g -fsanitize=address" clean all
src/tig 2>asan.log
```
repeat the same actions
```
READ of size 8 at 0x61d000002878 thread T0
    #0 0x474502 in file_finder_draw src/ui.c:143
    #1 0x4751ef in file_finder_input_handler src/ui.c:271
    #2 0x4328c8 in read_prompt_handler src/prompt.c:173
    #3 0x431fef in prompt_input src/prompt.c:67
    #4 0x432931 in read_prompt_incremental src/prompt.c:188
    #5 0x4755b9 in open_file_finder src/ui.c:302
    #6 0x45e2e7 in open_blob_view src/blob.c:40
[...]

PR #801, opened 2018-03-19
Don't allow pos->lineno to underflow in file search (#800)

proski:
If no text is found in the file search, pos->lineno would become `(unsigned long)-1` which would lead to invalid memory access in `file_finder_draw()`


jonas:
Thanks

PR #803, opened 2018-03-19
Fix arithmetic operation on void pointer

proski:


proski:
One reason is that I want to be consistent with the existing code. Just look for other occurrences of `malloc`, `calloc` and `realloc` - many of them have casts. Another reason is that I want the code to be C++ compatible. I know many C projects that support compilation with a C++ compiler - it catches many more warnings than any C compiler. tig is not C++ compilable, but it can be done.

ThomasAdam:
Well, the existing code should also be changed -- and I'd argue that's what you should do instead of this pull request.

As far as compatibility with C++ compilers go -- that's a red-herring.  I really struggle with projects which try to do this -- and AFAICT, I don't consider tig to be in that camp.

proski:
The cast has been removed, it's not essential for fixing the `-Wpointer-arith` warning.

jonas:
Thanks @proski ! 👍

PR #811, opened 2018-04-09
Fix incorrect time displayed in some views due to timezone differences

PaulChanHK:
The issues
1. if date-local is no, timezone information is missing, and time is displayed incorrectly.
   e.g. "2018-04-09 15:15 +0800" is displayed as "2018-04-09 15:15" while
        "2018-03-19 04:48 -0700" is displayed as "2018-03-19 04:48" while
2. if date-custom is yes and %z is set in date-format, timezone is displayed incorrectly as "+0000"

With this change,
1. if date-local is no, display timezone, with the format string "%Y-%m-%d %H:%M %z"
2. if date-custom is yes and %z is not set in date-format, force to display local time
3. handle "%z" in the format string, display actual timezone instead of "+0000"

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -183,6 +183,9 @@ mkdate(const struct time *time, enum date date, bool local, const char *custom_f
 	static char buf[SIZEOF_STR];
 	struct tm tm;
 	const char *format;
+	bool fmt_tz;
+	char *tz_str;
+	char tz_rp[5];
 
 	if (!date || !time || !time->sec)
 		return "";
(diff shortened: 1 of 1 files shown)


jonas:
This breaks several tests which unfortunately didn't got reported because of a bug in the `tools/travis.sh` script.
Would you mind rebasing on the latest `master`?

jonas:
Thanks a lot. I ended up fixing the tests and tweaking the formatting locally before merging.

PR #812, opened 2018-04-09
Change some default view options in tigrc, for main & blame views

PaulChanHK:
1. display line number in main/blame views
2. display commit id in main view


jonas:
> One does not simply change some default view options

Seriously though. The only one that makes sense to me is the change to the blame view.

jonas:
I ended up making the blame view have the same default columns as `git-blame`.

PR #813, opened 2018-04-09
Fix line numbers in grep view when scrolled

andebjor:
Disable the addition of number of scrolled lines to the line number in
"grep" mode, as the line numbers in this case actually are search
results, and not line numbers of e.g. a viewed file.


jonas:
Thank you for the fix.

PR #828, opened 2018-05-21
Make tig the process group leader and automatically clean child processes

rolandwalker:
Make tig the process-group leader and give it the foreground connection to the TTY.  This announces to the OS "I am the interactive program taking input".

Being process-group leader allows for a simple and robust idiom for cleaning all child processes at exit time: tig can signal HUP to the entire process group (which tig can be certain that it owns).  While tig doesn't intentionally keep very long running child processes, it does depend on child processes quite a bit, so this cleanup is just good practice.

As noted in `test/README.adoc`, this PR adds a minor issue: if the user sets `stty tostop` in their interactive shell, then tig's test suite will not be able to run.  Actually, it will run, but then constantly stop again as output is generated.  That is as intended for `stty tostop`: not to let multiple processes contend for writes to the terminal.  The same problem would occur with `tig-pick`. 

However 
 * `stty tostop` is not the default setting on any Unixlike so far as I know
 * the test suite and `tig-pick` are edge uses

If a bug report does come in regarding `tig-pick`, I'm sure there's a fix with some added complexity in either `tig` or `tig-pick`.

### DIFF
--- a/test/README.adoc
+++ b/test/README.adoc
@@ -11,6 +11,9 @@ test scripts as long as `PATH` is set to include the directories `src/`
 and `test/tools`. The latter directory is where the test helper
 libraries are located, the most important of which is `libtest.sh`.
 
+The test suite requires `stty -tostop` to be set in the running terminal,
+which is typically the default.
+
 Options
 -------
 
--- a/include/tig/display.h
+++ b/include/tig/display.h
@@ -43,6 +43,7 @@ extern unsigned int current_view;
 #define view_is_displayed(view) \
 	(view == display[0] || view == display[1])
 
+void init_tty(void);
 void init_display(void);
 void resize_display(void);
 void redraw_display(bool clear);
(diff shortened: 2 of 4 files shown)


test/README.adoc
```diff
@@ -11,6 +11,9 @@ test scripts as long as `PATH` is set to include the directories `src/`
 and `test/tools`. The latter directory is where the test helper
 libraries are located, the most important of which is `libtest.sh`.
 
+The test suite requires `stty -tostop` to be set in the running terminal,
+which is typically the default.
```
jonas: 👍

jonas:
Thanks.

rolandwalker:
Without knowing what vim does, there might be no easy fix for that, and you have to revert.

I am very familiar with how a daemonized process works and less familiar with an interactive process.

I didn't try to look at how tig responds to `SIGSTOP` and `SIGTSTP` here (relevant to `stty tostop` above, and most likely to your issue as well) because those are not normal signals.

The deepest motivation is really that the `hangup_children` idiom, in this form, is so simple and robust.  There are other less elegant ways to do the same thing.

#736 requires this, but is not important.

PR #832, opened 2018-05-24
Fix tig-pick output redirection

odnoletkov:
Current version doesn't work for my environment (bash bundled with macOS):

	$ sh --version
	GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin17)
	Copyright (C) 2007 Free Software Foundation, Inc.

The result commit is printed to stderr instead of stdout. I can't pinpoint the
problem but other traits in the script suggest that it is intended for bash
anyway. And it works fine with explicit bash for me.

### DIFF
--- a/contrib/tig-pick
+++ b/contrib/tig-pick
@@ -20,14 +20,14 @@ trap "rm -f '$CONFIG'" EXIT
 
 # Prepare config file: source user config, if present
 if [ ! -z "$TIGRC_USER" ]; then
-	echo "source $TIGRC_USER" >> $CONFIG
+	echo "source $TIGRC_USER" >> "$CONFIG"
 elif [ -f "$HOME/.tigrc" ]; then
-	echo "source $HOME/.tigrc" >> $CONFIG
+	echo "source $HOME/.tigrc" >> "$CONFIG"
 fi
 
 # Bind Enter to print the selected commit ID to error output and exit after
 # that.
-echo 'bind main <Enter> <sh -c "echo %(commit) >&2"' >> $CONFIG
+echo 'bind main <Enter> <sh -c "echo %(commit) >&2"' >> "$CONFIG"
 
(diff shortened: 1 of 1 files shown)


odnoletkov:
cc @kgraefe

kgraefe:
I don't think I tested the script with anything else than Bash. I did use a syntax checker to remove Bashisms, but idk, maybe that wasn't complete?

I'd be fine with that PR.

rolandwalker:
I don't understand why `tig-pick` doesn't simply use STDOUT.

You are right that there are some non-POSIX idioms in it.

Have you tried keeping `/bin/sh` and changing the non-POSIX `echo -n`?

```diff
diff --git i/contrib/tig-pick w/contrib/tig-pick
index e4789f303..d06f4023c 100755
--- i/contrib/tig-pick
+++ w/contrib/tig-pick
@@ -40,7 +40,7 @@ stderr=$(tig "$@" 3>&2 2>&1 1>&3 3>&-) || {
 commit=$(echo "$stderr" | tail -n1)

 # Check return value for valid commit ID
-if ! echo -n "$commit" | grep -iqE '^[0-9a-f]{40}$'; then
+if ! printf '%s' "$commit" | grep -iqE '^[0-9a-f]{40}$'; then
        echo "$stderr" >&2
        exit 1
 fi
```

odnoletkov:
@rolandwalker I've tried it and it does fix the issue as well. Turns out original problem was indeed `-n` not interpreted as an argument in sh-compatibility mode of my bash. Thanks!

I think it's better to stick with sh then so I've updated the pull request.

jonas:
LGTM, thanks for the fix.

kgraefe:
@odnoletkov I didn't know that bash had a sh-compatibility mode. Thanks for mentioning.

Issue #802, opened 2018-03-19
Tig diff view doesn't scroll using up and down keys, when in log view there's opened diff preview


GRiMe2D:
Steps to reproduce:
1. Launch tig
2. Log is shown, on <kbd>Enter</kbd> shows diff preview on right, and preview can be scrolled with <kbd>j</kbd> or <kbd>k</kbd> which is right.
3. Select some revision in log, and press <kbd>d</kbd> to open full diff window.
4. <kbd>Up</kbd> and <kbd>Down</kbd> keys switch previous or next revisions

If you don't open diff preview in step 2, 4th step work properly, <kbd>Up</kbd>, <kbd>Down</kbd> scrolls diff window

PR #835, opened 2018-05-27
Fix incorrect behaviour of up and down keys in diff view when opened from diff preview

koutcher:
I believe issue #802 is the main reason why so many people ask about remapping up and down keys. If the current behaviour is actually a feature maybe a parameter can be introduced.

### DIFF
--- a/src/tig.c
+++ b/src/tig.c
@@ -235,7 +235,7 @@ view_driver(struct view *view, enum request request)
 
 	case REQ_NEXT:
 	case REQ_PREVIOUS:
-		if (view->parent) {
+		if (view->parent && view == display[1]) {
 			int line;
 
 			view = view->parent;


src/tig.c
```diff
@@ -235,7 +235,7 @@ view_driver(struct view *view, enum request request)
 
 	case REQ_NEXT:
 	case REQ_PREVIOUS:
-		if (view->parent) {
+		if (view == display[1]) {
 			int line;
 
 			view = view->parent;
```
jonas: Maybe `view->parent && view == display [1]` so this won't default.

jonas:
Contextually speaking, I guess this makes more sense. I don't personally use the fullscreen navigation, but I remember doing it that way be because it felt more consistent at the time.

koutcher:
I personally switch to full screen only when a commit requires extra attention and I always end up using the the arrow keys to navigate as there is not the commit list to remind me that up and down keys are bound to this list.

I amended the PR to have view->parent && view == display [1].

jonas:
Thanks.

PR #836, opened 2018-05-29
Fix error when trying to open diff view from split view on (Un)staged changes line

koutcher:
The problem occurs when you have staged or unstaged changes in the repository and is mostly visible when diff-highlight is true. If you press d or enter in the main view it will automatically switch to stage view while if you press d from the split view it opens diff view and you get an annoying "Failed run the diff-highlight program: diff-highlight". For consistency, switch to stage view as it is done for the main view when there is no commit associated to the selected line.


jonas:
Thanks for fixing another usability issue Thomas. I've seen this but never got annoyed enough to investigate it further.

PR #837, opened 2018-05-31
Fix Vim going background after running Tig

koutcher:
Vim seems happier when we set the foreground process group ID back to its original value before leaving. Sounds logical when it's written like that actually.


jonas:
Sweet, this looks like an elegant and completely logical fix.

/cc @rolandwalker

jonas:
Thanks a lot.

Issue #845, opened 2018-06-27
Tig does not maximize after search


Perlence:
Tig stops resizing correctly after search. Here's a video https://youtu.be/001uQ35RYKI
I built tig on Debian 9.4. It happened to me both on `TERM=xterm-256color` and `TERM=tmux-256color`. Apparently this started happening after f21329ee4c69e20e76f4ffb2766b84a770644452.

koutcher:
The problem is not actually linked to the search function (any command you type will cause the same issue) but to the use of readline. This can be confirmed by using --without-readline option when compiling tig.

It is quite annoying but it looks like it won't be easy to solve: http://invisible-island.net/ncurses/ncurses.faq.html#readline_library

Perlence:
Thank you for investigating! Hm, running `./configure --without-readline` still leaves readline in config.h, so I had to comment it out manually. Well, I personally can live without readline, since the bug is much more annoying than being unable to use search history.

koutcher:
--without-readline works ok on macOS but is broken on Linux. You also have to remove readline from config.make to avoid linking with libreadline.so

Perlence:
Thanks, it's good to know!

jonas:
Fixed in #846

Perlence:
Good job y'all, thanks!

PR #846, opened 2018-06-29
Fix resize not working after entering command (#845)

koutcher:
Mininum Readline version is now 6.3 due to the use of the
rl_change_environment variable.

Also fixed check against major required version and added support for
configure --without-readline.

### DIFF
--- a/tools/ax_lib_readline.m4
+++ b/tools/ax_lib_readline.m4
@@ -211,21 +211,23 @@ AC_DEFUN([AX_LIB_READLINE], [
   	[search for readline in DIR/include and DIR/lib])],
 	[ac_cv_rl_prefix=$with_readline])
 
-  AC_CHECK_HEADERS([readline/readline.h], [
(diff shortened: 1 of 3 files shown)


tools/ax_lib_readline.m4
```diff
@@ -211,21 +211,23 @@ AC_DEFUN([AX_LIB_READLINE], [
   	[search for readline in DIR/include and DIR/lib])],
 	[ac_cv_rl_prefix=$with_readline])
 
-  AC_CHECK_HEADERS([readline/readline.h], [
-    AC_CHECK_HEADERS([readline/history.h], [
-      RL_LIB_READLINE_VERSION
+  AS_IF([test "x$with_readline" != xno], [
```
jonas: Are these changes made manually or did you update it to the newest version of this file?

koutcher: Manually.

Issue #787, opened 2018-03-02
branch not displayed when identically named tag exists


g-raud:
Tested with tig 2.2 and 2.3.0:
```
$ git branch temp
$ git tag temp HEAD^
$ git log # shows the branch temp
$ tig # does not show the branch temp
$ git checkout temp
warning: refname 'temp' is ambiguous.
Switched to branch 'temp'
$ tig # does not show the branch temp either
```

jonas:
Closed by #849

PR #849, opened 2018-07-07
Use fully qualified ref name for tag when conflicting with branch name

### LINKED ISSUE #746: Branch not visible, if there is a tag with the same name
If there are a tag and branch with identic names at one commit, only the tag is shown an the branch is missing.

koutcher:
Closes #746 and #787


jonas:
Thanks.

PR #851, opened 2018-07-10
Fix deleted file mode line remains highlighted after hovering in diff or stage view

koutcher:
Taking benefit of the half time break, here is one more patch. This was the last recurring annoyance I encounter with tig. Do you plan a new release soon ?


jonas:
Yes, I will release soon. Long overdue.
Thanks again for contributing bug fixes.

PR #864, opened 2018-08-10
Fix garbled cursor line with older ncurses versions

koutcher:


jonas:
Thanks, I didn't test this.

Issue #881, opened 2019-01-02
Support log.follow automatically


duncanmak:
`tig` has the `--follow` option (#361), but since Git 2.9, there's a git config option called `log.follow`.

    log.follow

    If true, git log will act as if the --follow option was used when a single <path> is 
    given. This has the same limitations as --follow, i.e. it cannot be used to follow 
    multiple files and does not work well on non-linear history.

If this is turned on, `tig` gets a bit confused, and the commits start veering off to the right.

It would be great if `--follow` can be turned on automatically if `log.follow` is enabled.

I'm running `tig version 2.4.1` on OS X.

duncanmak:
Thanks @jonas!

With this change, what happens if `log.follow` is enabled?

PR #882, opened 2019-01-02
Turn on '--follow' if 'log.follow' is set to true in git config

duncanmak:
This is a tiny patch that address #881

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -1463,7 +1463,8 @@ read_repo_config_option(char *name, size_t namelen, char *value, size_t valuelen
 			argv_append(&opt_log_options, "--pretty=medium");
 
 	} else if (!strcmp(name, "log.follow") && !strcmp(value, "true")) {
-		argv_append(&opt_rev_args, "--follow");
+		if (opt_file_args && !opt_file_args[1]) /* single file */
+			argv_append(&opt_rev_args, "--follow");
 	}
 
 	return SUCCESS;
--- a/src/options.c
+++ b/src/options.c
@@ -1461,6 +1461,9 @@ read_repo_config_option(char *name, size_t namelen, char *value, size_t valuelen
 	} else if (!strcmp(name, "format.pretty")) {
 		if (!prefixcmp(value, "format:") && strstr(value, "%C("))
 			argv_append(&opt_log_options, "--pretty=medium");
+
+	} else if (!strcmp(name, "log.follow") && !strcmp(value, "true")) {
+		argv_append(&opt_rev_args, "--follow");
 	}
 
 	return SUCCESS;


duncanmak:
I don't know why the AppVeyor build failed, the tests passed on my local machine. It'd be nice to retrigger a build, perhaps it's just a glitch.

koutcher:
`git log --follow` is only accepted with a pathspec, so, with this patch, any attempt to start tig without argument while log.follow is set to true in git config will fail. This is not reflected by the test suite because log.follow is never set, so this is not the reason for the AppVeyor build to fail.

duncanmak:
@koutcher Yes, you're totally right. I realized the same thing last night.

Do you know of a way to check that `tig` is being asked to load a `pathspec` and not a directory/repo?

koutcher:
I believe the test `opt_file_args && !opt_file_args[1]` should ensure there is one pathspec and only one. 

Though forcing —follow argument could do the job it would be interesting to understand why it works as the log.follow set to true in git config already makes git log to behave as git log —follow. There must be something else inside tig induced by the presence of the —follow which maybe could be trigged when log.follow is true.

duncanmak:
@koutcher Thanks for the tip and I have pushed a new commit. I believe the behavior is more correct now.

The reason why `--follow` is needed is because it turns off the graph rendering.  Without it, the graph will render and the lines veer off to the right.

With `log.follow` enabled (but no patch):

image

With `--follow` (no patch):

image
So this patch is simply to show the `--follow` behavior automatically when `log.follow` is turned on and we're viewing a single file.

P.S. Perhaps `with_graph` should always be `false` when we're only viewing a single file?

jonas:
Git has support for rewriting the history when filtering, this is enabled by the `git log --parents` flag:

```
/opt/tig@master > git log --parents src/tig.c | grep ^commit | head -3
commit a7eb2a3a8b188388c0eb2c6fe476990ad83c7fed 2272096ce3dccc3622c7f095a67005d405be2e9c
commit 2272096ce3dccc3622c7f095a67005d405be2e9c ccbda97bda65810cd416b5b4fbf7127c2e0aa951
commit ccbda97bda65810cd416b5b4fbf7127c2e0aa951 71e8361184fb19845e65fdd06538948a5376ccbf
/opt/tig@master > git log --follow --parents src/tig.c | grep ^commit | head -3
commit a7eb2a3a8b188388c0eb2c6fe476990ad83c7fed 2272096ce3dccc3622c7f095a67005d405be2e9c
commit 2272096ce3dccc3622c7f095a67005d405be2e9c d74e28a89959e3ed949bac32a21a5cc72de37554
commit ccbda97bda65810cd416b5b4fbf7127c2e0aa951 71e8361184fb19845e65fdd06538948a5376ccbf
/opt/tig@master > git -c log.follow=true log --parents src/tig.c | grep ^commit | head -3
commit a7eb2a3a8b188388c0eb2c6fe476990ad83c7fed 2272096ce3dccc3622c7f095a67005d405be2e9c
commit 2272096ce3dccc3622c7f095a67005d405be2e9c d74e28a89959e3ed949bac32a21a5cc72de37554
commit ccbda97bda65810cd416b5b4fbf7127c2e0aa951 71e8361184fb19845e65fdd06538948a5376ccbf
```
History rewriting is disabled when `--follow` is passed or `log.follow` is enabled and **a single path is given** according to the `git-config` man page:

> `log.follow`
>           If true, git log will act as if the --follow option was used when a single <path> is given. This has the same limitations as --follow, i.e. it cannot be used to follow multiple files and does not work well on non-linear history.

It sounds like we have to introduce a global variable for tracking the `log.follow` option and use it for automatically enabling/disabling the graph. This could be done in addition to the existing `no_graph` check in `src/argv.c:argv_parse_rev_flag`.

koutcher:
As suggested by @duncanmak in his P.S., maybe the graph does not make sense when pathspec is specified. Could be done just by adding `&& !opt_file_args` in src/main.c:main_with_graph. 

Using tig --author=... the other day, I wondered also if it shouldn't be without graph as well.

jonas:
History rewriting works for a single file, but I agree that a linear graph is not useful. I am okay with disabling the graph in this case.

And yes, it might make sense to disable the graph entirely for author searches. Another option would be to dynamically try to detect fragmented graphs and redraw them.

koutcher:
Disabling the graph when a pathspec is specified is finally too aggressive as history rewriting works quite well most of the time. The monorepo addicts will hate it. 

Switching to tracking the log.follow option to disable the graph.

PR #895, opened 2019-02-04
Git diff colors: interpret 'ul' as 'underline'

enizor:
# Why this change?
The official `git config` uses `ul` for `underline`.
It is this setting that is used by `diff-highlight` or `diff-so-fancy`.

The change allows for `.gitconfig` that works for both `tig` and `diff-highlight`

# Alternatives

There is probably a better manner or place to do it, I am not very fluent in C.

It could also be possible to use a separate `diff-add-highlight color color underline` in `.tigrc`. In this case `.tigrc` should take precedence over `.gitconfig` (this is not in effect currently).

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -1354,6 +1354,9 @@ parse_git_color_option(struct line_info *info, char *value)
 	for (i = 0; i < argc; i++) {
 		int attr = 0;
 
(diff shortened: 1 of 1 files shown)


jonas:
Thanks.

PR #896, opened 2019-02-08
Propagate --first-parent to diff arguments

### LINKED ISSUE #861: `tig --first-parent -m` does not show diffs like `git log -p --first-parent -m`
I use `git log -p --first-parent -m` to look at a repository that makes heavy use of merges, but of course, I'd much rather be using tig.  Unfortunately, tig doesn't show the diffs I expect, and I don't know if this is a bug or some unfortunate defaults that I have to work around.

For background, `git log`'s `-m` is:

```
-m
  This flag makes the merge commits show the full diff like regular commits;
  for each merge parent, a separate log entry and diff is generated. An exception
  is that only diff against the first parent is shown when --first-parent option
  is given; in that case, the output represents the changes the merge brought
  into the then-current branch.
```

Here's a repro case:

```
git clone https://github.com/NixOS/nixpkgs
cd nixpkgs
git checkout 117680b0e9f4472e6161427ac0814fc2c6f998ba
tig --first-parent -m
```
and press Enter on the commit.

For https://github.com/NixOS/nixpkgs/commit/117680b0e9f4472e6161427ac0814fc2c6f998ba, `tig --first-parent -m` shows a ~27662 line diff, while `git log -p --first-parent -m` shows the ~45 line diff I expect.

```
# tig --version
tig version 2.4.0
ncursesw version 6.1.20180127
readline version 6.3

# git --version
git version 2.18.0
```

koutcher:
Closes #861

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -231,8 +231,10 @@ main_check_argv(struct view *view, const char *argv[])
 			continue;
 		}
 
-		if (!strcmp(arg, "--first-parent"))
+		if (!strcmp(arg, "--first-parent")) {
 			state->first_parent = true;
+			argv_append(&opt_diff_options, arg);
+		}
 
 		if (!argv_parse_rev_flag(arg, &rev_flags))
 			continue;


jonas:
I have lots count of all these valuable fixes. Super greatful.

PR #904, opened 2019-03-20
Add synchronous command description in tigrc

typebrook:
Simply add missing description of  `+` prefix for user-defined command which is in man page


jonas:
Thanks.

PR #907, opened 2019-04-05
Don't shift signed int by 31 bits

proski:
cppcheck reports an issue in tig:

[view.h:75]:
(error) Shifting signed 32-bit value by 31 bits is undefined behaviour

There is no good reason to make VIEW_RESET_DISPLAY the largest value in
enum. Just use 1 << 17 for it, which is the next available power of two.


jonas:
I think I gave it that value because it felt a bit "kludgy"

PR #908, opened 2019-04-05
Fix incorrect cppcheck warning about realloc() use

proski:
cppcheck reports:
[view.c:1679]:
(error) Common realloc mistake: 'box' nulled but not freed upon failure

Indeed, writing the result of realloc() to its first argument is a bad
practice. It is OK here only because box and line->data are equal when
realloc() is called. Even if box becomes NULL, line->data survives.

To avoid the confusion, initialize box with the result of realloc(). Use
line->data until that point.


jonas:
👍 Thanks

PR #909, opened 2019-04-06
Untracked only status view

koutcher:
Following discussions with @proski in #762, I decided to give a try to a "Untracked only" status view when called from the Untracked changes line. Having in mind the use case of checking quickly which files were left behind, this version seems indeed better than opening the regular status view and moving the cursor to the Untracked files currently implemented.

@jonas, the refresh part surely needs to be reviewed.

### DIFF
--- a/src/status.c
+++ b/src/status.c
@@ -160,11 +160,15 @@ error_out:
 			watch_apply(&view->watch, WATCH_INDEX_STAGED_NO);
 		else if (type == LINE_STAT_UNSTAGED)
 			watch_apply(&view->watch, WATCH_INDEX_UNSTAGED_NO);
+		else if (type == LINE_STAT_UNTRACKED)
+			watch_apply(&view->watch, WATCH_INDEX_UNTRACKED_NO);
 	} else {
 		if (type == LINE_STAT_STAGED)
 			watch_apply(&view->watch, WATCH_INDEX_STAGED_YES);
 		else if (type == LINE_STAT_UNSTAGED)
 			watch_apply(&view->watch, WATCH_INDEX_UNSTAGED_YES);
+		else if (type == LINE_STAT_UNTRACKED)
+			watch_apply(&view->watch, WATCH_INDEX_UNTRACKED_YES);
 	}
 
 	io_done(&io);
--- a/src/status.c
+++ b/src/status.c
@@ -24,6 +24,7 @@
 #include "tig/git.h"
 #include "tig/watch.h"
 #include "tig/status.h"
+#include "tig/main.h"
 #include "tig/stage.h"
 
 /*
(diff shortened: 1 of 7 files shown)


src/status.c
```diff
@@ -738,6 +746,8 @@ status_request(struct view *view, enum request request, struct line *line)
 		return request;
 	}
 
+	if (view->parent)
```
koutcher: These 2 lines are problematic. When present, if you open the Untracked only status view from the Untracked files line, stage one file, open the regular status view (with s key) and then stage another file then the refresh is not correct anymore. @jonas, any idea how to achieve the automatic refresh of the parent view properly ?

src/status.c
```diff
@@ -31,6 +31,14 @@
  */
 
 static char status_onbranch[SIZEOF_STR];
+static bool show_untracked_only=false;
+
+void
+open_status_view(struct view *prev, bool untracked_only, enum open_flags flags)
+{
+	show_untracked_only=untracked_only;
+	open_view(prev, &status_view, flags | OPEN_REFRESH);
```
jonas: Is the `OPEN_REFRESH` serving a purpose? It was not there in the inline version.

src/status.c
jonas: Trigger a reload of the main view? Some views like stage has explicit logic based on the parent view if I recall correctly.

src/status.c
koutcher: Yes, without it, if you open the status view from the Untracked changes line and then want to open it after with the s key to have the full content, then the content remains the Untracked only.

src/status.c
koutcher: I noticed it and I also noticed there is no direct call to update_view, but I could not spot which part is the trigger for this logic.

koutcher: I found it ! update_view is actually called for parent in main_status_exists itself called by stage_exists at the end of stage_request. So my code should be correct.

src/status.c
jonas: It looks like it causes an assertion failure.

koutcher:
OK, I see the assert(flags ^ OPEN_REFRESH)... I have to try with OPEN_RELOAD instead.

proski:
I've tested the PR, the behavior is totally reasonable, fits well with the rest of tig.

test/main/untracked-test
```diff
@@ -53,15 +53,15 @@ $YYY_MM_DD_HH_MM +0000 Unknown    o Staged changes
...
 
-[status] Press u to stage all files for addition                            100%
+
+
+
+
+
+
+
+[status] Nothing to update                                                  100%
```
jonas: Sweet

src/status.c
```diff
@@ -738,6 +754,9 @@ status_request(struct view *view, enum request request, struct line *line)
 		return request;
 	}
 
+	if (show_untracked_only && view->parent && !main_status_exists(view->parent, LINE_STAT_UNTRACKED))
+		return REQ_VIEW_CLOSE;
```
jonas: Even if the status can only ever have the main view as the parent it would still be good to have a `view->parent == &main_view` before calling `main_status_exists`.

koutcher: Wilco

koutcher:
I change the view->parent test to view->parent == &main_view, rebased on master and squashed a bit.

src/status.c
jonas: I like "Impossible Germany" :-)
https://www.youtube.com/watch?v=r_0BPqWpM_M

jonas:
I'll let you merge.

PR #911, opened 2019-04-11
only accept "-C" before a subcommand

peff:
Since commit ed36f75 (Add -C option to specify the working directory, 2019-03-21), we allow:

```
tig -C/path/to/repo
```

to change into the repo directory before running any git commands. But because of the way tig's command-line parser works, this eats any "-C" we meant to pass to a git command, like:

```
tig blame -C foo.c
```

which used to enable copy-detection in the blame. Let's require that -C come before any sub-command. That will continue to allow:

```
tig -C/path/to/repo
```

with no subcommand. It also means that using "-C" with a subcommand looks like this:

```
tig -C/path/to/repo blame foo.c
```

which matches git's global -C option to change directories. And of course it allows:

```
tig blame -C foo.c
```

to work as it used to prior to ed36f75, enabling copy detection.

### DIFF
--- a/src/tig.c
+++ b/src/tig.c
@@ -515,6 +515,7 @@ parse_options(int argc, const char *argv[], bool pager_mode)
 
 	} else {
 		subcommand = NULL;
+		i--; /* revisit option in loop below */
 	}
 
 	for (; i < argc; i++) {
--- a/src/tig.c
+++ b/src/tig.c
@@ -475,10 +475,22 @@ parse_options(int argc, const char *argv[], bool pager_mode)
 
 	request = pager_mode ? REQ_VIEW_PAGER : REQ_VIEW_MAIN;
 
-	if (argc <= 1)
+	/* Options that must come before any subcommand. */
+	for (i = 1; i < argc; i++) {
+		const char *opt = argv[i];
+		if (!strncmp(opt, "-C", 2)) {
+			if (chdir(opt + 2))
+				die("Failed to change directory to %s", opt + 2);
+			continue;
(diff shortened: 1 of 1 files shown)


jonas:
Thanks Jeff. I will take a look at fixing the tests and get this merged ASAP.

peff:
Thanks. Let me know if I can help on the tests. I got a ton of failures running `make test` on master, so I didn't explore whether this might be breaking something new.

peff:
Looking at the Travis failures, I think there's a sort of off-by-one in my commit, fixed in https://github.com/jonas/tig/pull/911/commits/45a782124c26892241db4478384a7ebdb95d20cd.

src/tig.c
```diff
@@ -475,10 +475,22 @@ parse_options(int argc, const char *argv[], bool pager_mode)
...
+		} else {
+			break;
+		}
+	}
+
+	if (i >= argc)
 		return request;
 
-	subcommand = argv[1];
+	subcommand = argv[i++];
```
jonas: The subcommand is optional so the `++` here is what I think is causing problems.

```c
int i;

//...
for (i = 1; i < argc; i++) {
        // parse pre-arg: -C
}

subcommand = argv[i];
// parse subcommand

if (subcommand)
        i++;

for (; i < argc; i++) {
        // parse remaining args
}
```

peff: Yep, I think our comments crossed. Should be fixed by the commit I just pushed up.

jonas:
Thanks

PR #914, opened 2019-04-11
make-builtin-config: use "read -r"

peff:
When we bake the tigrc into the binary during the build process, the data passes through a shell `while read` pipeline. Even though we take care to add an extra layer of backslash escaping to any backslashes in the source material, the behavior of a backslash at the end of a line does not seem portable between shells:

```
$ bash tools/make-builtin-config.sh | grep git-colors
	"set git-colors = \\\n"

$ dash tools/make-builtin-config.sh | grep git-colors
	"set git-colors =\n"
```

It looks like `dash` treats the backslash before the newline as significant even when it's escaped. That yields a builtin config that doesn't parse (because the value lines for git-colors which should be continuations appear as their own badly-formatted config keys).

We can fix this by using `read -r`, which is the POSIX-approved way of inhibiting backslash handling. And of course we have to strip out the extra layer of backslash quoting in sed, which is now unnecessary (we still have to double all of the backslashes to protect them from sed itself).


peff:
After this PR I can almost run `make test` to completion on my Debian unstable box. The exception is that the output from one of the emoji test results omits the `\u{fe0f}` (variant-16) modifier to some of the emojis. I have no idea why this is the case, and I gave up on digging.

jonas:
Thanks. For the emojis, I must admit it is all black magic to me at this point with ncurses and the terminal having seemingly different ideas of how to handle them.

PR #915, opened 2019-04-13
new feature: import/export marked commit lists

aaptel:
When doing large backports or similar work that involves a lot of
cherry picking, it can be useful to maintain lists of commits. While
in the main view, Tig can mark commits via the toggle-bp-mark action,
which is bound to C-b by default. Marked commits title will be
displayed using an alternative style (main-bp-mark).

If a path is specified when running Tig (see `-l` option). Tig can
read and write marked commits. Tig tries to import that file on
startup if it exists, and writes to that file on exit.

The format of this file is simple, each line must have the following format:

    <sha1>[ <optional text>]

Tig tries to keep the content of each line when it imports
it. Additional commits added during the Tig session will use the
commit title as text when exporting.

On exit, Tig sorts the marked commits by commit date (not author date)
and exports the marked lists to the file specified by the `-l` option.

---

* new toggle-bp-mark user request (C-b)
* new main-bp-mark font style
* new bplist.{c,h} module to manipulate instances of bplists
* new run opt `-l` to specify location to import/export marked list to

Currently tig has one global bplist, but the implementation is generic
enough and works on bplist instances so that in the future we could
have multiple bplists.

A bplist holds a list of lines and a hashtable that maps fully-expanded revs to lines.

* bplist_init(): initialize a bplist
* bplist_read(): load a bplist from a file into an initialized bplist
* bplist_has_rev(): checks whether a bplist contains a commit rev
* bplist_add_rev(): adds a rev to a bplist if it is not already there
* bplist_rem_rev(): removes a rev from a bplist if it holds it
* bplist_toggle_rev(): adds/remove rev from a bplist
* bplist_write(): dump a bplist to a file

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -312,6 +312,7 @@ TIG_OBJS = \
 	src/grep.o \
 	src/ui.o \
 	src/apps.o \
+	src/bplist.o \
 	$(GRAPH_OBJS) \
 	$(COMPAT_OBJS)
 
--- a/doc/manual.adoc
+++ b/doc/manual.adoc
@@ -237,6 +237,33 @@ be appended:
 [main] 77d9e40fbcea3238015aea403e06f61542df9a31 - commit 1 of 779 (0%) 5s
 -----------------------------------------------------------------------------
 
+[[bp-mark-list]]
+Commit Marks
+------------
+
+When doing large backports or similar work that involves a lot of
+cherry picking, it can be useful to maintain lists of commits. While
+in the main view, Tig can mark commits via the toggle-bp-mark action,
+which is bound to <C-b> by default. Marked commits title will be
+displayed using an alternative style (main-bp-mark).
+
+If a path is specified when running Tig (see `-l` option). Tig can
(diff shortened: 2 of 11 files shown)


magnayn:
I came here just to say this is exactly the feature I'm looking for!

My dev workflow is somewhat messy, and tends to be
a) pull master
b) interactive rebase, where I 
  select a number of commits and bubble them to the top
c) push several commits down the head to master

At the moment a is cmdline, c I actually use tig (previously to get the sha1, but now I found I can usefully bind a command), but (b) is a fairly rote 'git rebase -i'.

I think someone else was after a 'mark this commit so I can trigger a diff against it with a different commit' which would also be of interest.

koutcher:
An alternative to this PR could be (with tig master):
* Add `bind generic    <C-B>   @sh -c "git notes --ref=bplist add -m 'To be cherry picked' %(commit) || git notes --ref=bplist remove %(commit)"` to .tigrc
* Start `tig --notes=bplist` or `tig --notes=*`
* Toggle mark with C-b

PR #925, opened 2019-05-07
 Add support for repos created with git --work-tree 

koutcher:


src/blame.c
```diff
@@ -77,7 +77,7 @@ static enum status_code
 blame_open(struct view *view, enum open_flags flags)
 {
 	struct blame_state *state = view->private;
-	const char *file_argv[] = { repo.cdup, view->env->file , NULL };
+	const char *file_argv[] = { repo.is_inside_work_tree ? repo.cdup : repo.git_dir, view->env->file , NULL };
 	char path[SIZEOF_STR];
```
jonas: Given that this code is repeated many places, it might merit a dedicated member in the struct or alternatively a macro: `repo_exec_dir(repo)` or `repo.exec_dir`.

koutcher: Agreed. Introduced `repo.exec_dir`.

Issue #809, opened 2018-04-06
Feature Request: Ability to add comments into help view.


MironVeryanskiy:
**_Feature Request :_** Ability to add comments into the help view.

**Reason**: There is no way to provide comments for custom keybinds in the help view.

![image](https://user-images.githubusercontent.com/1906941/38440035-943e7260-39a5-11e8-9ef2-ecf6b6bd6ae8.png)

**Attempt:** I tried to add hacky comments using shell in-line execution, but the # character still gets exposed to .tigrc parser, and stripped away.

![image](https://user-images.githubusercontent.com/1906941/38439919-095bef92-39a5-11e8-9bd9-36a8704c11ab.png)

PR #926, opened 2019-05-18
Add comments for bind commands [#809]

stevenyvr987:
A user may have defined numerous and complex key bindings. The user will show the help screen to be reminded of the key bindings. The user will see keybindings for built-in actions accompanied by help text, because they are built into the source code. However, the user will see user-defined key bindings without help text, because any comments typed in the configuration file are stripped when parsed. Without seeing the comments, the user may not remember the meaning of a key binding from seeing just the action name.

The PR adds code to enable comments in bind options to be shown in the help screen. For example, `bind keymap key action # comments` will result in `comments` displayed next to `action`, in the same way that help text for built-in actions are shown. Also, if a / search is done, the PR adds code to ensure that comment text are searched.

There are two exceptions.

1. If a `bind` is defined in a repo configuration file, `tig` will first strip comments by using `git config --list` before parsing the `bind` option. Consequently, comments can't be shown in the help screen for `bind` options appearing in repo config files.

1. If a `bind` is defined on the prompt line, `tig` will not parsed any comments introduced by a hash mark, because it's not expected the user will type any. Consequently, comments will be parsed as command parameters.

The PR is organized into several commits. See details in the various commit messages.

1. A prepatory commit. Add a new function to argv.c similar to `argv_to_string_alloc()`, but it adds a prefix to the allocated text buffer. In the process, a bug is discovered and fixed in `argv_to_string_alloc()`: if a non-empty separator is specified, `argc` will fail to count the addition of separators. Since existing calls of the function always specify empty separators, there has been no impact from the bug.

1. Add code to show the comment text in the help screen. At this stage, comment text is shown as free-form text and not aligned with built-in help text.

1. Add code to align comment text when shown in the help screen; the code uses the function introduced in the first commit. Only keybindings with comment text will participate in calculating the field width; keybindings with no comments but with long action names may overflow the field.

1. The expected output of two existing test cases have changed and have been corrected. A modest test case is added for the new behaviour.

1. Make sure that new comment text for keybindings can participate in grep searches, ie, if the user initiates a / search on the prompt line, matches in comment text should show up. Unfortunately, I can't come up with a unit test for this.

1. Simplify the code a bit and trade a bit of space for time by caching key strings during `help_open()` so that `help_draw()` doesn't need to recalculate them when drawing a help line.

I can follow up with another commit to make the necessary changes to documentation.

### DIFF
--- a/include/tig/argv.h
+++ b/include/tig/argv.h
@@ -24,6 +24,7 @@
 #define DIFF_ARGS "%(diffargs)"
 
 bool argv_to_string(const char *argv[], char *buf, size_t buflen, const char *sep);
+char *argv_to_string_alloc_prefix(const char *argv[], const char *sep, const char *prefix);
 char *argv_to_string_alloc(const char *argv[], const char *sep);
 bool argv_to_string_quoted(const char *argv[SIZEOF_ARG], char *buf, size_t buflen, const char *sep);
 bool argv_from_string_no_quotes(const char *argv[SIZEOF_ARG], int *argc, char *cmd);
--- a/include/tig/keys.h
+++ b/include/tig/keys.h
@@ -90,6 +90,7 @@ struct run_request {
 	struct keymap *keymap;
 	struct run_request_flags flags;
 	const char **argv;
+	char *name;
 	char *help;
 };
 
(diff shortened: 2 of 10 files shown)


stevenyvr987:
Looks like I missed the case where a user executes a bind option on the prompt line.
Will fix and update the PR.

stevenyvr987:
Fixed; force-pushed PR branch; updated my original comment to the PR.

stevenyvr987:
Another update of the PR. I've added a new parameter to `set_option()`, which allows me to pass in comment text in a simpler way.

PR #929, opened 2019-06-01
Update INSTALL - improve installation w/ configure

sarpik:
I was building from the github repo and found it weird that the `configure` file was not present.
Only randomly I happened to notice that the `make configure` is what generates it.

I know I should've read the whole thing, but sometimes you might not have enough time and/or you just want things to work, thus I made this change.

(I tried moving up the `Note` part, but it looked kinda out-of-place, so I guess you could experiment with it yourself so that it's not annoying for those who build from the release but those who build from source see it before actually building).

### DIFF
--- a/INSTALL.adoc
+++ b/INSTALL.adoc
@@ -43,6 +43,7 @@ Installation using `configure`
 
 Optionally, you can use the `configure` script to detect dependencies:
 
+	$ make configure	# if you're building from the Tig repository (requires `autoconf`)
 	$ ./configure
 	$ make
 	$ make install
--- a/INSTALL.adoc
+++ b/INSTALL.adoc
@@ -43,7 +43,7 @@ Installation using `configure`
 
 Optionally, you can use the `configure` script to detect dependencies:
 
-	$ make configure	# if you're building from the Tig repository (requires `autoconf`)
+	$ ./autogen.sh	    # if you're building from the Tig repository (requires `autoconf`)
 	$ ./configure
 	$ make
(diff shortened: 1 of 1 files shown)


jtheoof:
There is an `autogen.sh` to generate the file. Seems pretty standard when using `autoconf`. See [the doc](https://www.gnu.org/software/automake/faq/autotools-faq.html#What-does-_002e_002fbootstrap-or-_002e_002fautogen_002esh-do_003f)

INSTALL.adoc
```diff
@@ -43,6 +43,7 @@ Installation using `configure`
 
 Optionally, you can use the `configure` script to detect dependencies:
 
+	$ make configure	# if you're building from the Tig repository (requires `autoconf`)
```
jtheoof: `./autogen.sh`

sarpik: Alright, thanks for the info @jtheoof. Should I change the line to `$ ./autogen.sh` or do nothing and just close this PR?

jtheoof: No I would keep the MR but using my recommended fix.

Perhaps @jonas will merge. It's fairly easy to review.

sarpik: If this is what you've meant - I've changed it to `./autogen.sh`.

jonas:
I ended up moving the note about running `make configure` up. Maybe it won't get read as you said. I will think about whether to split the section into "building with configure from git source" and "build from release tarball".

Thanks for the PR!

PR #934, opened 2019-06-11
Introduce reflog view

### LINKED ISSUE #538: [RFE] add `git reflog` information to tig
It would be great if I could see the reflog in tig and browse those commits. I use `git reflog` [sometimes](https://kparal.wordpress.com/2011/09/01/git-tip-of-the-day-recover-lost-commits/) to recover commits that no longer are being pointed to by some other ref.

If this feature is already there and I missed it, I apologize.

koutcher:
Press L to open the reflog view or start with tig reflog.

The main benefit over git reflog --pretty=raw | tig --pretty=raw
is the ability to browse the full chain of commits for each reflog.

Closes #538

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -299,6 +299,7 @@ TIG_OBJS = \
 	src/watch.o \
 	src/pager.o \
 	src/log.o \
+	src/reflog.o \
 	src/diff.o \
 	src/help.o \
 	src/tree.o \
--- a/doc/manual.adoc
+++ b/doc/manual.adoc
@@ -96,6 +96,10 @@ The log view::
 	Presents a more rich view of the revision log showing the whole log
 	message and the diffstat.
 
+The reflog view::
+	Presents a view of the reflog allowing to navigate the repo
+	history.
+
 The diff view::
 	Shows either the diff of the current working tree, that is, what
 	has changed since the last commit, or the commit diff complete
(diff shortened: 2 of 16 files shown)


gipsy-king:
Cool, I was just looking for this in the issues and found this here. Tried locally, it works!

Would it make sense to add `bind reflog    C       ?git checkout %(branch)` to `tigrc`? That is my main use case, but I don't know for what purpose other people use `git reflog` mainly. Then also `%(branch)` is not ideal, when going through rebases etc., you might want to checkout a commit instead. But there doesn't seem to be a `branch or fallback to commit` available in `tigrc`.

koutcher:
It is quite unlikely that you want to do a `checkout %(branch)` from the reflog because if your commit is reachable with an actual ref, you shouldn't need the reflog. 

The most typical use cases would be `cherry-pick` to recover a single commit or `reset` to undo a rebase for example. 

There is already a binding for cherry pick in the main view and I added a `!` binding to the reflog view for `reset --hard`.

gipsy-king:
Sounds good. I can always add it to my `tigrc`.

koutcher:
I figured out your use case, `bind reflog C ?git checkout %(branch)` added to tigrc.

gipsy-king:
Wow, thanks ;)

PR #942, opened 2019-07-04
Properly detect combined diffs

### LINKED ISSUE #941: Viewing merges with --first-parent gives "Allocation Failure"
Happens with current master (faf1b865e08a26437d2d42e5b0e45f8dea2239bf).
Does not happen in 2.4.1 (though a diff is also not shown).

I bisected this to 909b4e5, I think it might be because `state->combined_diff` is set incorrectly?

In the previous commit (21f3829e87d2f8edc0afe75e32affc644222707a) it works and correctly shows the non-combined diff as expected for `--first-parent`.

moben:
Combined diff chunks start with '@@@', use that to detect them instead of
the "Merge:" field that is present for all merges, regardless of diff type.

Technically, this still doesn't handle merges with more than 2 parents
but that's not a regression

From git-log(1):

     4. Chunk header format is modified to prevent people from
        accidentally feeding it to patch -p1. Combined diff format was
        created for review of merge commit changes, and was not meant
        for apply. The change is similar to the change in the extended
        index header:

            @@@ <from-file-range> <from-file-range> <to-file-range> @@@

        There are (number of parents + 1) @ characters in the chunk
        header for combined diff format.

Closes #941

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -247,7 +247,9 @@ diff_common_read(struct view *view, const char *data, struct diff_state *state)
 	if (!view->lines && type != LINE_COMMIT)
 		state->reading_diff_stat = true;
 
-	if (state->combined_diff && !state->after_diff && data[0] == ' ' && data[1] != ' ')
+	/* combined diffs lack LINE_DIFF_START and we don't know
+	 * if this is a combined diff until we see a "@@@" */
+	if (!state->after_diff && data[0] == ' ' && data[1] != ' ')
 		state->reading_diff_stat = true;
 
 	if (state->reading_diff_stat) {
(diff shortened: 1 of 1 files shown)


src/diff.c
```diff
@@ -296,6 +292,8 @@ diff_common_read(struct view *view, const char *data, struct diff_state *state)
 		return true;
 
 	} else if (type == LINE_PP_MERGE) {
+		/* we will check again when highlighting diff chunks but we
```
koutcher: Thanks for spotting and fixing this issue. What about removing this part as you initially proposed and changing the test `if (state->combined_diff && !state->after_diff && data[0] == ' ' && data[1] != ' ')` to `if (!state->after_diff && data[0] == ' ' && data[1] != ' ')` as the check on after_diff should be sufficient here ? Maybe it would be wise to add a comment stating that this test is necessary because merge commits lack the LINE_DIFF_START.

moben: I agree, that's better.

moben: Thanks for the review!

PR #945, opened 2019-07-13
 Fix #631 crossing branch lines have the wrong color

### LINKED ISSUE #631: Crossing branch lines have the wrong color
When branch lines are crossing each other, the horizontal line has the color of the branch it is crossing (the one going vertically), instead of being the same color that continues after the crossing. It is especially apparent if the line crosses multiple, differently colored vertical lines. See the following screenshots. In both cases, the horizontal lines should be purple.

screenshot from 2017-06-30 15-13-55

screenshot from 2017-06-30 15-14-14

stevenyvr987:
Issue #631 notes that

> When branch lines are crossing each other, the horizontal line has the color of the branch it is crossing (the one going vertically), instead of being the same color that continues after the crossing. It is especially apparent if the line crosses multiple, differently colored vertical lines.

The mis-coloring occurs because of three factors:
1. A crossover symbol shows vertical and horizontal lines, for example, "─│" for UTF-8 line-graphics.
2. A symbol can have one color assignment.
3. Graphs are rendered with vertical lines taking priority over horizontal lines, ie, the vertical line will have the correct color while the horizontal line will be mis-colored.

An obvious fix is to simplify the crossover symbol by removing the mis-colored horizontal component, for example, use " │" for"'─│", but the result is unsatisfying because the path in which the horizontal line was part of becomes too broken.

The chosen fix is to remove the vertical component instead, ie, use "──" for "─│", and then use the same color as the right-hand neighbour. Another crossover symbol to fix is "─╭"; here we remove the horizontal component and simplify the symbol to " ╭". The resultant vertical path extend further on the screen, and so to the eye, it does not break as much as the horizontal path. A graph using horizontal crossover symbols is quite satisfying and may actually be preferred.

We introduce a new toggle option for `commit-title-graph` to enable the user to switch between the old and the new ways. If the command `set main-view-commit-title-graph = v2-horizontal-crossover` is used instead of the default `set main-view-commit-title-graph = v2`, then the cross-over symbols will give priority to horizontal paths and as a result there will not be mis-colorings.

The PR is organized into three commits.
1. The main commit; code changes are summarized in the commit message.
2. Update the relevant test suite.
3. Update the tigrc man page.

### DIFF
--- a/doc/tigrc.5.adoc
+++ b/doc/tigrc.5.adoc
@@ -460,8 +460,13 @@ author::
 	  sizing content, up to this limit.
 
 commit-title::
-	- 'graph' (mixed) [no|v2|v1]: Whether to show the revision graph in the
-	  main view on start-up. "v1" refers to the old graph rendering, which
+	- 'graph' (mixed) [no|v2|v2-horizontal-crossover|v1]:
+	  Whether to show the revision graph in the main view on start-up.
+	  "v2" refers to the newer graph rendering where a vertical path is placed over a horizontal path,
+	  but the horizontal path is miscolored.
+	  "v2-horizontal-crossover" refers to the newer graph rendering where a horizontal path is placed over a vertical path,
+	  and the horizontal path is correctly colored.
+	  "v1" refers to the old graph rendering, which
 	  is less accurate but faster and thus recommended in large
 	  repositories.  See also the 'line-graphics' options.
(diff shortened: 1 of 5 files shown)


alerque:
Any feedback from maintainers on this? I happened to notice the same bug and then found the issues and now I see a fix has been pending for 5 years with no feedback (at least not visible here) as to why it is or is not able to move forward.

Issue #795, opened 2018-03-16
tab completion is really slow


dwijnand:
in a clone of https://github.com/scala/scala run `tig blame src/library/scala/math/Order<TAB>` and (on my machine) it'll take 14 seconds to give you:

```
$ tig blame src/library/scala/math/Order
src/library/scala/math/Ordered.scala    src/library/scala/math/Ordering.scala
```

tig version 2.3.3
git version 2.15.1

pzrq:
@dwijnand Great question. Could you run `$ time tig blame src/library/scala/math/Order` and copy/paste the output here? 

If you can think of any other way to debug, profile or otherwise get visibility into what is causing the slowness that would be great too.

jonas:
The bash tab completion is still based on a very old copy of the git
completion. Could be part of the problem.

>

dwijnand:
> Could you run `$ time tig blame src/library/scala/math/Order<TAB>` and copy/paste the output here?

you can't time the bash completion - it just outputs the same thing:

```bash
$ time tig blame src/library/scala/math/Order
src/library/scala/math/Ordered.scala    src/library/scala/math/Ordering.scala
```

pzrq:
Then is this a tig issue or a bash issue or some interaction of the two?

dwijnand:
it's an issue with tig's bash completion.

for comparison completing `blame src/library/scala/math/Order<TAB>` is less than a second for me.

askedrelic:
I've noticed this as well on a git repo with 14K tags. I'm assuming tags are part of the autocomplete algorithm and that is what I assume really slows it down. I think it's more due to bash being slow at large amounts of data.

tig version 2.3.3
git version 2.16.3

ingydotnet:
On a huge repo I use daily, `tig` Bash completion hangs for 2 minutes. If I type `tig ` and press tab, I'm screwed. Git completions on same repo are fast.

Here's a fun trick:
```
set -x; tig <tab><tab>
```
Try that on a big repo and sit back with some coffee and watch the show.

rohieb:
The culprit is probably __tig_complete_file:

```
linux.git $ time __tig_complete_file 

real    0m29.363s
user    0m27.805s
sys     0m1.600s

linux.git $ time __git_complete_file

real    0m0.014s
user    0m0.003s
sys     0m0.008s
```

[...]

Issue #940, opened 2019-06-28
zsh:12: command not found: __tig_main


g4rick:
![image](https://user-images.githubusercontent.com/4426450/60331712-defa5480-99c7-11e9-9818-117970d5dda1.png)

It happened When I input `tig` and then tap ‘tab’ button.
```bash
zsh:12: command not found: __tig_main
```

Some info:

* I install tig use homebrew. 
* use zsh & oh-my-zsh with init .zshrc file (only add nvm)

```bash
tig --version

tig version 2.4.1
ncurses version 5.7.20081102
readline version 8.0

git --version

git version 2.22.0
```

Garonenur:
I have/had the same problem.

I installed the completion file from this repository to ~/.completion but it did not work.

reading in the logs of `zsh -lx` I saw that some `_git` completion file was loaded and defined `_tig` I found this in `/usr/share/zsh/functions/Completion/Unix/_git` which is probably different for you, but it might hint in the right direction.
I renamed that file so it gets ignored by zsh-completion and now the completion file that comes with this repo is used as advertised.

koutcher:
@rohieb proposed an updated version of the completion for tig in #960, could you give it a try with zsh ?

ahmedre:
tried the updated version of completion and it didn't help - but likely not because of that PR itself, since the issue seems to be coming from something else.

as @Garonenur mentions, if I remove the `_git` file located in `/usr/local/share/zsh/site-functions/_git`(this is where Homebrew installs it), then completion works as expected.

i suspect the reason has to do with the bottom of `_git`, where, within the `_git()` method, it says:

```
        if (( $+functions[__${service}_zsh_main] )); then
                __${service}_zsh_main
        else
                emulate ksh -c __${service}_main
        fi
```

the `emulate ksh` line is what causes it to look for `__tig_main` (as evidenced by commenting out that line no longer giving the same error, but also not doing any completion for `tig`). i am not sure why pressing tab on `tig` kicks off the `_git()` method though.

Garonenur:
our workaround is not working for me anymore, I removed the _git file that came with ubuntu (WSL) and I have git and tig compiled from source, the respective completion files are in my ```$fpath``` (I use zsh).
If I have the system installed ```_git``` I get the above mentioned error, ```__tig_main``` not found. If I remove that ```_git``` file, and thus only have the one from gits github source I now get ```tig-completion.bash:98: command not found: __git_complete``` the first time I try completion and 
```
[...]

PR #960, opened 2019-10-17
bash/zsh completion: reimplement and decrease runtime by factor 1863

rohieb:
Fixes: #795 ("tab completion is really slow")

When run on a Git repository with a large amount of files, like the linux repository, __tig_complete_file() can take a very long time to complete. This is unfortunate because when accidentally pressing Tab in such a repo, bash gets stuck until the completion function has finished, and does not even allow the user to cancel the operation with Ctrl-C.

In contrast, the current git completion does not have these problems, and since tig's command line parameters are mostly parameters to git-log or git-diff, we can use git's native completion for those cases instead. This also has the advantage that we do not need to care about updating the tig completion when new parameters are added to git-log or git-diff.

I have tested this only in bash, not on zsh, so it would be nice to get some input from zsh users.

For comparison, here is an exemplary runtime measurement of the old and the new completion in bash, showing an improvement of factor 1863. Admittedly, this was on a fairly loaded system (a build server), but still then the new completion runs in unnoticable time. I'm also getting similar results on an idle system in the same repo with runtime improvements of about factor 1000.

    linux (master) $ echo $BASH_VERSION
    5.0.3(1)-release

    linux (master) $ git describe
    v5.4-rc3-38-gbc88f85c6c09

    linux (master) $ uptime
     16:45:52 up 36 days,  3:33, 224 users,  load average: 24.17, 38.87, 31.21

    # The old completion:
    linux (master) $ . /usr/share/bash-completion/completions/tig
    linux (master) $ time COMP_WORDS=("tig log") COMP_CWORD=2 _tig

    real    2m1.145s
    user    1m40.379s
    sys     0m1.347s

    # The new completion:
    linux (master) $ . ../tig/contrib/tig-completion.bash
    linux (master) $ time COMP_WORDS=("tig log") COMP_CWORD=2 __git_wrap_tig

    real    0m0.127s
    user    0m0.085s
    sys     0m0.024s

With this change, almost nothing of the old completion remains, so change the copyright header accordingly. I'm also now adding a GPL-2.0-or-later dedication, which is the same license as most other code in this repository, and which, I presume, was also the author's intent since the first incarnation of this file.

### DIFF
--- a/contrib/tig-completion.bash
+++ b/contrib/tig-completion.bash
@@ -1,276 +1,88 @@
-##
-# bash completion support for tig
-#
+# bash/zsh completion for tig
+# 
+# Copyright (C) 2019 Roland Hieber, Pengutronix
 # Copyright (C) 2007-2010 Jonas fonseca
-# Copyright (C) 2006,2007 Shawn Pearce
 #
-# Based git's git-completion.sh: http://repo.or.cz/w/git/fastimport.git
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of
+# the License, or (at your option) any later version.
 #
-# The contained completion routines provide support for completing:
-#
-#    *) local and remote branch names
-#    *) local and remote tag names
-#    *) tig 'subcommands'
-#    *) tree paths within 'ref:path/to/file' expressions
-#
+# This completion builds upon the git completion (>= git 1.17.11),
+# which most tig users should already have available at this point.
 # To use these routines:
 #
-#    1) Copy this file to somewhere (e.g. ~/.tig-completion.sh).
-#    2) Added the following line to your .bashrc:
-#        source ~/.tig-completion.sh
+#    1) Copy this file to somewhere (e.g. ~/.bash_completion.d/tig).
+#
+#    2) Add the following line to your .bashrc:
+#
+#           source ~/.bash_completion.d/tig
+#
+#       Note that most Linux distributions source everything in
+#       ~/.bash_completion.d/ automatically at bash startup, so you
+#       have to source this script manually only in shells that were
+#       already running before.
 #
 #    3) You may want to make sure the git executable is available
 #       in your PATH before this script is sourced, as some caching
 #       is performed while the script loads.  If git isn't found
 #       at source time then all lookups will be done on demand,
 #       which may be slightly slower.
-#
 
-__tigdir ()
-{
-	if [ -z "$1" ]; then
-		if [ -n "$__git_dir" ]; then
-			echo "$__git_dir"
-		elif [ -d .git ]; then
-			echo .git
-		else
-			git rev-parse --git-dir 2>/dev/null
-		fi
-	elif [ -d "$1/.git" ]; then
-		echo "$1/.git"
-	else
-		echo "$1"
-	fi
-}
+__tig_options="
+	-v --version
+	-h --help
+	-C
+	--
+	+
+"
+__tig_commands="
+	blame
+	grep
+	log
+	reflog
+	refs
+	stash
+	status
+	show
+"
 
-_tigcomp ()
-{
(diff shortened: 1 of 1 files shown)


markrian:
This is *great*, @rohieb, thank you!

After copying this to `/usr/share/bash-completion/completions/tig`, I'm finding that `__git_complete` isn't defined in a given terminal session until I've tried to complete a `git` command. That is, I need to attempt to complete a `git` command _before_ I attempt to complete a `tig` command; otherwise nothing happens.

An example from a new terminal session:

```sh
-bash: __git_complete: command not found

$ complete | grep "\(git\|tig\)"
complete -o bashdefault -o default -F _fzf_path_completion git

$ git show mas<tab><tab> ^C # this is where I try to complete a git command

$ complete | grep "\(git\|tig\)"
complete -o bashdefault -o default -o nospace -F __git_wrap__gitk_main gitk
complete -o bashdefault -o default -o nospace -F _fzf_path_completion git

```
I have no idea how bash completions are wired up, but it seems weird to me that things that are hooked up with `complete` change after trying to complete commands!

Is there something I'm missing, here?

koutcher:
@hohieb, could you add the missing tig `-C` argument and `reflog` command (completion should be identical to `tig log`) and rebase ?

rohieb:
@koutcher ah, thanks for the note. `-C` was a bit tricky because of the missing space before the path, until I found the `compgen -P` option. Learned something along the way! :D

I also updated the install comments and made a few other minor style and typo fixes.

koutcher:
Great. One problem though, If you do `tig -C<Tab>`, make a completion and then `<Space><Tab>`, bash goes into infinite loop. The cause seems to be the `continue` that was added line 57 to `-*) ;;`.
Also, `tig reflog` does not take the same argument as `git reflog`. It is more like `tig log`, so I believe we should have:
```patch
diff --git a/contrib/tig-completion.bash b/contrib/tig-completion.bash
index e7756d9..355b15f 100755
--- a/contrib/tig-completion.bash
+++ b/contrib/tig-completion.bash
@@ -73,6 +73,9 @@ _tig() {
 		refs|status|stash)
 			COMPREPLY=( $(compgen -W "$__tig_options" -- "$cur") )
 			;;
+		reflog)
+			__git_complete_command log
+			;;
 		"")
 			__git_complete_command log
 			__gitcompappend "$(compgen -W "$__tig_options $__tig_commands" -- "$cur")"
```

rohieb:
> If you do `tig -C<Tab>`, make a completion and then `<Space><Tab>`, bash goes into infinite loop.

Mh, of course I forgot to test this case at all, or else I would've noticed 😳… Should be fixed now.

> Also, `tig reflog` does not take the same argument as `git reflog`. It is more like `tig log`

Thanks for the clarification, I now did it as per your suggestion.

fjarlq:
According to @spazm in this comprehensive stackoverflow answer, this pull request created a regression in zsh completion:

https://stackoverflow.com/a/64164918/393684

fjarlq:
@felipec Are you saying [you think you know how to fix this problem?](https://stackoverflow.com/questions/62766815/git-completion-in-zsh-git-func-wrap3-not-found#comment114074643_64164918)

rohieb:
@fjarlq it might be related to #940 too

felipec:
@fjarlq My guess is that [this commit](https://github.com/felipec/git-completion/commit/d7c7498378e8201d5f5e6eca621600ca9f78abb6) might fix the issue.

felipec:
Actually, no, the issue is that bashcompinit was loaded at the wrong time.

I fixed that [here](https://github.com/felipec/tig/commit/604f80fca9966d34e4f4c92c43b7348f8f732723).

But also, there's a bunch of other issues. I'm sending a pull request.

darcyparker:
@markrian FYI: In https://github.com/jonas/tig/pull/960#issuecomment-574253386, the part labeled "Previous version of comment" has some details that could be explained by some of my notes in https://github.com/jonas/tig/issues/1011#issuecomment-736922515.

Issue #47, opened 2012-04-30
color-words support


Oblomov:
Currently, tig has no support for color-words in diffs. In fact, starting tig wit the --color-words option fails with "No revisions match the given argument" (which is quite misleading as error message).

This should probably be consiered a long-term TODO, as it requires some heavy-duty re-engineering of the diff view, especially with respect to highlighting. Also, color-word support would require massaging the input parameters, since the word-diff option should always be turned into 'porcelain'.

wilzbach:
Any news on this? It's been open for more than four years now ;-)

timotheecour:
+1

hhartz:
attempted exporting `TIG_DIFF_OPTS=--color-words` but no luck, seems inline color escapes are not supported?

+1

SassNinja:
+1 for this feature

I've just discovered tig but without `--color-words` it's much less valuable for me than it would be with it

avatar-lavventura:
Any improvement in this feature?

hydrargyrum:
If I'm not mistaken, it is implemented in tig 2.5.5. I added this to my tigrc:

```
bind diff D :toggle word-diff
```

and it works like a charm when pressing `shift-d` (the separators are a bit verbose, but this is git's `plain` mode, not tig's fault).

trygveaa:
> the separators are a bit verbose, but this is git's plain mode, not tig's fault

I don't think you can argue that it's git's fault when git has a color mode for the word diff. I, and I would assume most people, would like this way to display it, not the plain way.

hydrargyrum:
Maybe the selected word-diff mode should be configurable in tig then. Plain mode is still useful for example if the terminal has no color support.

PR #966, opened 2019-11-11
Added word-diff=plain colorizing support

### LINKED ISSUE #221: --word-diff
Hi, I've just discover the amazing --world-diff options, which diff by world instead of line.
It works with git show and git diff.
I would like to be my default diff method, but I can't make it work.
I  tried to set TIG_DIFF_OPTS but it doesn't work.

Also, when I try
tig show --word-diff, it doesn't display the colors :-(

dvalter:
Added ADD and DEL colors for {+added+} and [-removed-] blocks of git word-diff=plain

Since git does not escape delimiters in word-diff, this code may produce an incorrect color output on some code that actually contains these sequences including the first commit of this PR.

It's possible to bypass this by using git color escape sequences to determine additions and deletions, but it's likely to conflict with the rest of diff parser.

(Probably) fixes #221

### DIFF
--- a/include/tig/line.h
+++ b/include/tig/line.h
@@ -94,7 +94,8 @@ struct ref;
 	_(PALETTE_12,		""), \
 	_(PALETTE_13,		""), \
 	_(GRAPH_COMMIT,		""), \
-	_(SEARCH_RESULT,	"")
+	_(SEARCH_RESULT,	""), \
+	_(WDIFF_PLAIN,		"")
 
 enum line_type {
 #define DEFINE_LINE_ENUM(type, line) LINE_##type
--- a/src/diff.c
+++ b/src/diff.c
@@ -276,11 +276,15 @@ diff_common_read_diff_wdiff(struct view *view, const char *text)
 	 *	there could be [-one-] diff part{+s+} in the {+any +} line
 	 */
 
-	while (diff_common_read_diff_wdiff_group(&context));
+	while (diff_common_read_diff_wdiff_group(&context)) {}
 
+    // It's not a word-diff line, continue parsing
+    if (context.text == text)
+        return NULL;
+    
 	diff_common_add_cell(&context, strlen(context.text), false);
-
-	return diff_common_add_line(view, text, LINE_DEFAULT, &context);
+    
+	return diff_common_add_line(view, text, LINE_WDIFF_PLAIN, &context);
 }
 
 static bool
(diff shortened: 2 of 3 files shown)


eMPee584:
.. hey @koutcher, can you please have a look at this? :cowboy_hat_face:

koutcher:
The proper way to do this would involve `--word-diff=porcelain` (see #47) but I understand it would require much more rework of the parser, so this is an interesting step already. There is a problem when hovering the cursor line over empty lines (j,k) as they remain highlighted.

dvalter:
@koutcher I definitely missed that during my previous testing. It should be fixed now.

koutcher:
What is the purpose of LINE_WDIFF_PLAIN ?

dvalter:
I feel like since the line is parsed and colorized, it's not *default* anymore. Now it should not have an impact, but it may help in further development.
If that's not right, I'll remove LINE_WDIFF_PLAIN.

koutcher:
The line type is normally attached to a color line in tigrc so it wouldn't bring anything here. There was a better way to solve the cursor line problem (just change the last parameter of diff_common_add_cell to true) so I merged 9b9e6f2 and did a few adjustments to ensure it will be  transparent for people not using word diff. You can have a look to https://github.com/koutcher/tig/commit/41763ae3c1516004ccc32da79d64363a86a861be before I push it to jonas/tig.

dvalter:
> The line type is normally attached to a color line in tigrc so it wouldn't bring anything here

There's an alternative use case regarding key handling (i.e. DIFF_STAT), so I thought it may make sense if some action for diff add/del will be added.

>You can have a look to koutcher@41763ae before I push it to jonas/tig.

It's absolutely fine. I love that you've added a word-diff CLI argument switch, it was definitely needed there.

koutcher:
Point taken for the line type, we can introduce later is there is a need for a specific action. Thanks for this contribution!

vivien:
When I use `tig --word-diff=plain`, the changes are shown like `[-classic-]{+strict+}`. Am I missing something?

dvalter:
> When I use `tig --word-diff=plain`, the changes are shown like `[-classic-]{+strict+}`. Am I missing something?

It's an intended behavior, if they have DEL and ADD colors respectively. It should look basically like an output of `git show --word-diff=plain`.
Screenshot_2020-06-10_22-14-42
Screenshot_2020-06-10_22-15-35

Stripping them is not viable as it will break commits like 4028a495425.

PR #970, opened 2019-11-24
Add TIGRC_EXTRA

psprint:
I'm thinking about creating a plugin for Zsh, but also possibly for Bash, that would supply e.g.: a bindings that would display a fzf-backed list of Makefile targets and run a selected target. To supply such additional functionality in a pluggable fashion such variable would serve very well.

The patch is very simple. It loads the extra configuration after the 2 other ones.


koutcher:
I believe this can already be achieved using TIG_SCRIPT as highlighted in cb4f848.

PR #983, opened 2020-01-13
Fix vertical lines in Putty with UTF-8 graphics

### LINKED ISSUE #981: Vertical window separator renders as "x" in Putty, but the graph is OK
When run inside Putty, tig renders the vertical window separator in putty using `x` characters. However, the version graph uses correct vertical lines.

Observed with different fonts: Consolas, DejaVu Sans Mono, Fira Mono, Fixedsys, Inconsolata, Lucida Console, Lucida Sans Typewriter. Observed with putty 0.70 and 0.73. Tig 2.4.1 and the current git versions are affected:

```
$ /usr/bin/tig --version
tig version 2.4.1
ncursesw version 5.9.20140118
readline version 6.3
$ src/tig --version
tig version 2.5.0-2-gca0809d
ncursesw version 5.9.20140118
readline version 6.3
```

The system running Putty is Windows 10, the system running tig is Ubuntu 14.04.

The described behavior corresponds to `diff-view-line-graphics = utf-8`. If it's `ascii`, the vertical bar is rendered using `|`, but the graph becomes less pretty. For `default`, `x` is used for the vertical bar, and the graph looks even worse than in the `ascii` mode. The graph is rendered using letters that don't even look like lines.

```
 o make-builtin-config:
 Mqk Merge pull request
 x o Add default bindin
 x o Add option to star
 oqj Introduce reflog v
 o Fix ls-tree tree-ish
```

I suspect that the vertical window separator is rendered using the `default` mode even though the `utf-8` mode is in effect. Apparently, putty has an issue with `diff-view-line-graphics = default`.

I would consider this issue resolved if the window separator renders in the `utf-8` mode correctly.

putty-tig

proski:
Putty doesn't show VT100 line drawing characters in the UTF-8 mode by
   default. Putty 0.70 and older doesn't even have that option.
   
   For the vertical display separator, use '|' as background, then output
   the vertical line character on every line.
   
   For the line number separator, use draw_graphic() in the default mode
   only. Use draw_chars() in the ASCII and UTF-8 modes.
   
   Add a test for vertical lines. The display separator is always captured
   as a line of '|' characters. The line number separators are captured
   correctly. 'x' corresponds to ACS_VLINE.

   Closes #981

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -187,11 +187,24 @@ static void
 redraw_display_separator(bool clear)
 {
 	if (display_sep) {
-		chtype separator = opt_line_graphics ? ACS_VLINE : '|';
+		int lineno = 0;
 
 		if (clear)
 			wclear(display_sep);
-		wbkgd(display_sep, separator + get_line_attr(NULL, LINE_TITLE_BLUR));
+		wbkgdset(display_sep, get_line_attr(NULL, LINE_TITLE_BLUR));
+
+		switch (opt_line_graphics) {
+		case GRAPHIC_ASCII:
+			while (mvwaddch(display_sep, lineno++, 0, '|') == OK);
+			break;
+		case GRAPHIC_DEFAULT:
+			while (mvwaddch(display_sep, lineno++, 0, ACS_VLINE) == OK);
+			break;
+		case GRAPHIC_UTF_8:
+			while (mvwaddstr(display_sep, lineno++, 0, "│") == OK);
+			break;
+		}
+
 		wnoutrefresh(display_sep);
 	}
 }
--- a/src/draw.c
+++ b/src/draw.c
@@ -314,7 +314,7 @@ draw_lineno_custom(struct view *view, struct view_column *column, unsigned int l
 	unsigned long digits3 = MIN(9,MAX(3,column->width));
 	int max = MIN(VIEW_MAX_LEN(view), digits3);
 	char *text = NULL;
-	chtype separator = opt_line_graphics ? ACS_VLINE : '|';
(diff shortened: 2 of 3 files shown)


proski:
Fixed an issue with the vertical window separator not redrawing correctly after cycling graphics modes with `~` on Ubuntu 14.04. Using a simple one-line approach to iterate over lines without using `getmaxyx()` (_"it's better to ask forgiveness than permission"_).

proski:
* Using `switch` without `default`. If a new graphics mode is used, the compiler would warn about a unhandled option, which is a good thing.
* Made the line number separator code closer to the graph output code. ASCII and UTF-8 output strings, DEFAULT draws symbols.

proski:
* In the line number separator code, return `false` for the unsupported graphics setting to fix a compiler error.
* In the display separator code, use pipe (`|`) as background in the UTF-8 mode. It gets overwritten anyway, but the logic becomes simpler.

proski:
Added a test for vertical lines in all graphic modes. Unfortunately, the vertical display separator is always output as `|` symbols by `save-display`. However, the line number separators are captured correctly.

proski:
Changed `redraw_display_separator()` to be more uniform. The separator is now drawn in a loop in all graphic modes. `wbkgd()` is replaced with `wbkgdset()`, so the curses library can save some time clearing the window.

Using background to draw the separator is a nice trick, but it won't work in the UTF-8 mode without using `cchar_t` and introducing compatibility issues. It's better to do the same thing in all modes unless there are reasons to do it differently.

jonas:
Thanks a lot. Really great to have the test.

Issue #523, opened 2016-09-15
Crash when splitting chunk


wezm:
With `tig` in the following state, pressing `\` to call `stage-split-chunk` causes a crash for me every time.

![screen shot 2016-09-15 at 4 18 45 pm](https://cloud.githubusercontent.com/assets/21787/18540214/9748ab4c-7b60-11e6-8c75-bbf0d9a3bf3e.png)

The error/backtrace reported is:

```
** Error in `tig': realloc(): invalid next size: 0x000000000216af60 ***
======= Backtrace: =========
/usr/lib/libc.so.6(+0x70c4b)[0x7f4c8809ac4b]
/usr/lib/libc.so.6(+0x76fe6)[0x7f4c880a0fe6]
/usr/lib/libc.so.6(+0x79fcc)[0x7f4c880a3fcc]
/usr/lib/libc.so.6(realloc+0x159)[0x7f4c880a54f9]
tig[0x40602e]
tig[0x418db3]
tig[0x419018]
tig[0x4190cc]
tig[0x421908]
tig[0x4223dd]
tig[0x40371a]
/usr/lib/libc.so.6(__libc_start_main+0xf1)[0x7f4c8804a291]
tig[0x40400a]
======= Memory map: ========
00400000-0043a000 r-xp 00000000 08:02 3823205                            /usr/bin/tig
00639000-0063a000 r--p 00039000 08:02 3823205                            /usr/bin/tig
0063a000-00647000 rw-p 0003a000 08:02 3823205                            /usr/bin/tig
00647000-00650000 rw-p 00000000 00:00 0
01ec7000-021e9000 rw-p 00000000 00:00 0                                  [heap]
7f4c83de9000-7f4c83dff000 r-xp 00000000 08:02 3804638                    /usr/lib/libgcc_s.so.1
7f4c83dff000-7f4c83ffe000 ---p 00016000 08:02 3804638                    /usr/lib/libgcc_s.so.1
7f4c83ffe000-7f4c83fff000 r--p 00015000 08:02 3804638                    /usr/lib/libgcc_s.so.1
7f4c83fff000-7f4c84000000 rw-p 00016000 08:02 3804638                    /usr/lib/libgcc_s.so.1
7f4c84000000-7f4c84021000 rw-p 00000000 00:00 0
7f4c84021000-7f4c88000000 ---p 00000000 00:00 0
[...]

PR #990, opened 2020-02-08
Fix two memory issues

xaizek:
* Crash on diff chunk splitting (#523).
 * Memory leak on diffing.


koutcher:
Good catch. Thanks for sharing.

PR #1009, opened 2020-04-29
Blame uncommitted deletions starting from HEAD

### LINKED ISSUE #1008: Blame an unstaged deletion
I often want to find the commit that introduced a line which is deleted in the working tree or index.
Mostly for finding the right target for a fixup.

[Fugitive](https://github.com/tpope/vim-fugitive/) has spoiled me in this regard because I can press `o` on a deleted line followed by `:Gblame` to see which commit added the line.

For Tig, I think it would make sense to press `b` on a not-yet-committed line deletion in the diff view to do the same. Currently those lines will not show up in the blame view because `view-blame` does a blame with respect to the working tree. I wrote [this patch](https://github.com/krobelus/tig/commit/blame-from-head) to make the blame start at `HEAD`, which still has the lines that I deleted later. The starting cursor position is usually not too far from it, but can't be accurate in general because line numbers don't match the ones in the work tree.

I'm wondering if this is a reasonable change, or if there is maybe another way to do this.

<s>I think the proper fix would consist of invoking `git blame` once to find out which commit deleted the line at the cursor, and then showing the blame view with that commit and the correct line number.</s>

krobelus:
When starting a blame in the stage view on a deleted line, it seems
appropriate to find the commit that added the line.  To this end, instead
of starting the blame view from the working-tree-state of a file, use
the version at HEAD, which still has the deleted line (unless the line
was only just staged).  Compute the position of the deleted line to
start the blame view there.

Closes #1008

### DIFF
--- a/src/stage.c
+++ b/src/stage.c
@@ -420,7 +420,7 @@ stage_request(struct view *view, enum request request, struct line *line)
 		if (stage_line_type == LINE_STAT_UNTRACKED) {
 			open_editor(stage_status.new.name, (line - view->line) + 1);
 		} else {
-			open_editor(stage_status.new.name, diff_get_lineno(view, line));
+			open_editor(stage_status.new.name, diff_get_lineno(view, line, false));
 		}
 		break;
 
@@ -445,7 +445,7 @@ stage_request(struct view *view, enum request request, struct line *line)
 		}
 
 		view->env->ref[0] = 0;
-		view->env->goto_lineno = diff_get_lineno(view, line);
+		view->env->goto_lineno = diff_get_lineno(view, line, false);
 		if (view->env->goto_lineno > 0)
 			view->env->goto_lineno--;
 		return request;
--- a/src/stage.c
+++ b/src/stage.c
@@ -366,6 +366,118 @@ stage_chunk_is_wrapped(struct view *view, struct line *line)
 	return false;
 }
 
+bool
+find_deleted_line_in_head(struct view *view, struct line *line) {
+	struct io io;
+	struct buffer buffer;
+	unsigned long line_number_in_head, line_number = 0;
+	long bias_by_staged_changes = 0;
+	char buf[SIZEOF_STR] = "";
+	char file_in_head_pathspec[sizeof("HEAD:") + SIZEOF_STR],
+		file_in_index_pathspec[sizeof(":") + SIZEOF_STR];
+	const char *file_in_head = NULL;
+	const char *ls_tree_argv[] = {
+		"git", "ls-tree", "-z", "HEAD", view->env->file, NULL
+	};
+	const char *diff_argv[] = {
+		"git", "diff", "--root", file_in_head_pathspec, file_in_index_pathspec,
+		"--no-color", NULL
+	};
+
+	if (line->type != LINE_DIFF_DEL)
+		return false;
+
+	// Check if the file exists in HEAD.
+	io_run_buf(ls_tree_argv, buf, sizeof(buf), repo.exec_dir, false);
+	if (buf[0]) {
+		file_in_head = view->env->file;
+	} else { // The file might might be renamed in the index. Find its old name.
+		struct status file_status;
+		const char *diff_index_argv[] = {
+			"git", "diff-index", "--root", "--cached", "-C",
+			"--diff-filter=ACR", "-z", "HEAD", NULL
+		};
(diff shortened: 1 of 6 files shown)


src/stage.c
```diff
@@ -444,8 +444,13 @@ stage_request(struct view *view, enum request request, struct line *line)
 				string_ncopy(view->env->file, file, strlen(file));
 		}
 
-		view->env->ref[0] = 0;
-		view->env->goto_lineno = diff_get_lineno(view, line);
+		if (!is_initial_commit() && line->type == LINE_DIFF_DEL) {
```
koutcher, line 447: The reason for `!is_initial_commit()` here is not obvious to me, can you explain ?

krobelus: Oh well, in this case the blame will fail anyway with "Failed to load blame data". I removed it now, this is clearer.

koutcher: Ok I see what you wanted to do. What we'd need here is a way to detect an "AM" status...

koutcher: I quite agree that diff_get_old_lineno should be merged into diff_get_lineno. Something like `unsigned int diff_get_lineno(struct view *view, struct line *line, bool old)`.  Then, with stage.c modified this way, it should do it.

```patch
@@ -445,7 +445,16 @@ stage_request(struct view *view, enum request request, struct line *line)
 		}
 
 		view->env->ref[0] = 0;
-		view->env->goto_lineno = diff_get_lineno(view, line);
+		if (line->type == LINE_DIFF_DEL) {
+			const char *ls_tree_argv[] = {
+				"git", "ls-tree", "-z", "HEAD", view->env->file, NULL
+			};
+			char buf[SIZEOF_STR] = "";
+			io_run_buf(ls_tree_argv, buf, sizeof(buf), NULL, false);
+			if (buf[0])
+				string_copy(view->env->ref, "HEAD");
+		}
+		view->env->goto_lineno = diff_get_lineno(view, line, !!view->env->ref[0]);
 		if (view->env->goto_lineno > 0)
 			view->env->goto_lineno--;
 		return request;
```

krobelus: Yeah, that works nicely! Now the behavior for "AM" files is as before: the working-tree version of the file is opened in the blame-view. Not very useful/common but still better than just showing an error.

I've also added proper handling of "MM" files now - when blaming an unstaged deletion, I parse `git diff-index --cached -p HEAD -- filename` to compute the correct line number in `HEAD`.

What is still missing are uncommitted renames - blame currently has the old behavior here.
Perhaps the way to fix this is by running `git diff-index --cached -C -p HEAD` and then looking for a diff that renames the current file, though it would be nicer to reuse the status information here.

koutcher:
Couldn't `view->env->file` be passed as an argument to `git diff-index` to avoid parsing all the diffs ? Also, you may return a pointer on a local variable in find_deleted_line_in_head. Maybe it could be changed to `bool find_deleted_line_in_head(struct view *view, struct line *line, char *filename, unsigned long *lineno)` and `stage_request()` changes could be reduced to: 
```patch
+               if (find_deleted_line_in_head(view, line, view->env->file, &view->env->goto_lineno))
+                       string_copy(view->env->ref, "HEAD");
+               else
                        view->env->goto_lineno = diff_get_lineno(view, line, false);
                if (view->env->goto_lineno > 0)
                        view->env->goto_lineno--;
```

krobelus:
> Couldn't view->env->file be passed as an argument to git diff-index to avoid parsing all the diffs ?

This is to deal with RM files; If file `x` is renamed to `y` in the index, then `git diff-index HEAD y` doesn't know about the rename. I think there is no way around computing the diff of the entire tree to detect renames.

However, dealing with renames here is not super important to me, maybe we should leave that out for now?
Alternatively, I can do the entire diff (to find the renaming) only for those cases that actually need it: where `view->env->file` exists in the index but not in HEAD.

koutcher:
The use case is fine, I just didn't realize that providing a pathspec to `diff-index` was defeating the rename detection.

krobelus:
Addressed all comments.
I added an ls-tree call that allows to skip rename detection most of the time - when a file with the same name exists in HEAD, this will always use that filename Technically there can be false positives, but rename detection is only a heuristic, too.

koutcher:
Looks good to me. As Steve used to say, one more thing: so far, @jonas has maintained compatibility with pre-C99 compilers. Can you keep variables declaration at the beginning of blocks ? Also can you check NEWS.adoc, Improvements are listed before Bug fixes.

krobelus:
Should be done now, thanks!

PR #1013, opened 2020-05-14
Fix parsing of three-way chunk headers with extra dashes

krobelus:
This fixes a rare case of chunk line numbers not being computed
correctly.  It happens when a chunk marking a conflict contains a "-"
in its description:

	@@@ -45,7 -45,10 +45,14 @@@ void foo() { // -dash

### DIFF
--- a/src/parse.c
+++ b/src/parse.c
@@ -213,8 +213,9 @@ parse_chunk_header(struct chunk_header *header, const char *line)
 	if (!prefixcmp(line, "@@ -"))
 		line += STRING_SIZE("@@ -") - 1;
 	else if (!prefixcmp(line, "@@@") &&
-		 (line = strrchr(line, '-')))
-		/* Stay at that '-'. */ ;
+		 (line = strstr(line, " @@@")))
+		while (*line != '-')
+			line--;
 	else
 		return false;
 


koutcher:
Good catch, but there could be more than 2 parents. Maybe something like this would do the trick:
```patch
--- a/src/parse.c
+++ b/src/parse.c
@@ -213,8 +213,9 @@ parse_chunk_header(struct chunk_header *header, const char *line)
        if (!prefixcmp(line, "@@ -"))
                line += STRING_SIZE("@@ -") - 1;
        else if (!prefixcmp(line, "@@@") &&
-                (line = strchr(line, '-')) && (line = strchr(line+1, '-')))
-               /* Stay at that '-'. */ ;
+                (line = strstr(line, " @@@")))
+               while (*line != '-')
+                       line--;
        else
                return false;
```

krobelus:
Yes, that's better, and more explicit. I don't know a tool that produces such diffs with more than two parents, but of course it is possible.

koutcher:
I don't use them either but MariaDB/server has plenty of them, for example. Just a result of `git merge` with several branches. It all depends on your workflow.

krobelus:
Ah right, I was trying to produce a merge conflict from an octopus merge, but git refuses to do conflicting merges with more than two parents. If there is no conflict, it works just fine.

jonas:
Thanks a lot. Really great fix.

PR #1016, opened 2020-06-09
Don't error if git tab completion is not available. Fixes #1011

### LINKED ISSUE #1011: [BUG] -bash: __git_complete: command not found
I'm trying to set up tig along with bash-completion on a new machine (Mac). 

I installed the following packages through Homebrew: 
`brew install git bash-completion bash-git-prompt tig`

Versions:
bash - 3.2.57
tig - 2.5.1
git - 2.20.1
bash-completion: 1.3
bash-git-prompt: 2.7.1

I also configured my `.bash_profile` as per the info documentation for each package. 

But, while bash is sourcing `/usr/local/etc/bash_completion`, I get the following error:
`-bash: __git_complete: command not found`

When I grep through my entire `/usr/local/etc/bash_completion.d` directory, I see that the only reference to `__git_complete` is in the `tig-completion.bash` directory.

If I'm not mistaken, the`/bash_completion.d/tig-completion.bash` directory is installed when brew installs tig, as per the following post-install message:
```
==> Caveats
A sample of the default configuration has been installed to:
  /usr/local/opt/tig/share/tig/examples/tigrc
to override the system-wide default configuration, copy the sample to:
  /usr/local/etc/tigrc

Bash completion has been installed to:
  /usr/local/etc/bash_completion.d

zsh completions and functions have been installed to:
  /usr/local/share/zsh/site-functions
==> Analytics
```

I couldn't find any other issues related to this (and I'm not sure if this is the right place to post this issue, so I apologize if it's not). I did find another project that seems to have identified and patched this issue [here](https://github.com/NixOS/nixpkgs-channels/commit/f57da8ef0a9c4e4f4f003a92058b07f70b90fe72).

Am I doing something wrong here? I have the exact same configuration with slightly older library versions on another machine and everything works fine, so I'm guessing something has changed in the later versions of one of these libraries that is now causing this issue.

If this is a known bug, is there a workaround or work in progress to address it?

solarkennedy:
Seems like tab completion should always be a nice-to have, so made it just exit silently if git tab completions were not available. If you want I can make it print a friendly error message.

As is, users get a pretty mysterious:

```
-bash: __git_complete: command not found
$
```

And it is a little hard to know where it came from.

Fixes #1011

### DIFF
--- a/contrib/tig-completion.bash
+++ b/contrib/tig-completion.bash
@@ -98,7 +98,7 @@ fi
 
 # we use internal git-completion functions (if available), so wrap _tig for all necessary
 # variables (like cword and prev) to be defined
-if type '__git_complete' 2>/dev/null | grep -q 'function'; then
+if [ "$(type -t __git_complete)" = function ]; then
 	__git_complete tig _tig
 
 	# The following are necessary only for Cygwin, and only are needed
--- a/contrib/tig-completion.bash
+++ b/contrib/tig-completion.bash
@@ -28,6 +28,9 @@
 #       is performed while the script loads.  If git isn't found
 #       at source time then all lookups will be done on demand,
 #       which may be slightly slower.
+#
+#    4) This completion file depends on git completion already being
+#       loaded. Make sure git-completion.bash happens first.
 
 __tig_options="
 	-v --version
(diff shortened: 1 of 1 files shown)


contrib/tig-completion.bash
```diff
@@ -93,13 +96,15 @@ if [ -n "$ZSH_VERSION" ]; then
 	bashcompinit
 fi
 
-# we use internal git-completion functions, so wrap _tig for all necessary
+# we use internal git-completion functions (if available), so wrap _tig for all necessary
 # variables (like cword and prev) to be defined
-__git_complete tig _tig 
+if type '__git_complete' 2>/dev/null | grep -q 'function'; then
```
iillyyaa, line 101: (minor) `type -t __git_complete` will output `function`, so you may want to simplify this to:
```suggestion
if [ "$(type -t __git_complete)" = function ]; then
```

contrib/tig-completion.bash
```diff
@@ -93,13 +96,15 @@ if [ -n "$ZSH_VERSION" ]; then
 	bashcompinit
 fi
 
-# we use internal git-completion functions, so wrap _tig for all necessary
+# we use internal git-completion functions (if available), so wrap _tig for all necessary
 # variables (like cword and prev) to be defined
-__git_complete tig _tig 
+if [ "$(type -t __git_complete)" = function ]; then
```
darcyparker, line 101: Nitpick... `if ! declare -f __git_complete &>/dev/null; then` is a better test for existence of function because it doesn't spawn shell with `$()`

darcyparker:
See https://github.com/jonas/tig/issues/1011#issuecomment-736922515. I did some analysis and found that sourcing `git` completions is alterative to exiting silently.

darcyparker:
https://github.com/jonas/tig/pull/1055 is an alternate solution that attempts to define `__git_complete` and exits silently if it cannot.

rohieb:
Yes, the approach in #1055 looks more reasonable to me than just silently doing nothing if the git completion is not loaded yet.

solarkennedy:
Cool. I'll go ahead and close. Hopefully someone with write access can merge?

PR #1023, opened 2020-07-13
Add troubleshooting information

koutcher:


TROUBLESHOOTING.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+Note that the ability to set the `LANG` variable to values such as
+`en_DE` with KDE does not infer their validity.
+
+ESCDELAY
+~~~~~~~~
+
+See ESCDELAY environment variable in ncurses(3x) to reduce the one
+second delay after pressing the escape key. Nowadays,
+`export ESCDELAY=100` should be more than enough.
```
jonas, line 77: TIL 🤩

jonas (approved):
LGTM, I don't know if it would make sense to link to this page from the README's bug reporting section.

TROUBLESHOOTING.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+LANG
+~~~~
+
+The `LANG` variable must be set in accordance with your terminal
+encoding settings. If the `LANG` variable is not properly set,
+accented characters will be distorted.
+
+To check that your locale is present in the list of available locales,
+use `locale -a | grep "$LANG"`.
```
krobelus, line 67: I get
```
$ locale -a
C
en_US.utf8
POSIX
$ echo $LANG
en_US.UTF-8
```

and I think it works fine, so I guess

```suggestion
use `locale -a | grep -i "$LANG"`.
```

TROUBLESHOOTING.adoc
koutcher: > LGTM, I don't know if it would make sense to link to this page from the README's bug reporting section.

Good idea. Is this appropriate ?
```patch
 Bugs and Feature Requests
 -------------------------
+Please check out the
+https://github.com/jonas/tig/blob/master/TROUBLESHOOTING.adoc[troubleshooting guide]
+if you are experiencing issues with Tig.
+
 Bugs and feature requests can be reported using the
```

TROUBLESHOOTING.adoc
koutcher: Thanks for reporting, looks like `utf8` is a Linux oddity. `-i` won't help because of the missing hyphen. So I'd rather reword to:
```patch
-To check that your locale is present in the list of available locales,
-use `locale -a | grep "$LANG"`.
+Use `locale -a` to check if your `LANG` variable is present in the
+list of available locales (on Linux, you may find that `UTF-8` has
+been substituted for `utf8` but they are equivalent).
```

TROUBLESHOOTING.adoc
jonas: Yes, or we could also add it to `book.md` to make it part of the documentation site and link to it "relatively" using

```asciidoc
link:TROUBLESHOOTING.{docext}[troubleshooting guide]
```

I can do that if you prefer in a future PR.

koutcher: I have pushed my current state to the PR, I'll be happy if you want to take the lead on it. Edit at your convenience.

TROUBLESHOOTING.adoc
krobelus: Sounds good, thanks!
systemd's `localectl list-locales` uses `en_US.UTF-8`, but of course that's not available elsewhere.

TROUBLESHOOTING.adoc
```diff
@@ -0,0 +1,92 @@
...
+
+Note that the ability to set the `LANG` variable to values such as
+`en_DE` with KDE does not infer their validity.
```
krobelus, line 73:
```suggestion
Note that the ability to set the `LANG` variable to values such as
`en_DE` with KDE does not imply their validity.
```

TROUBLESHOOTING.adoc
```diff
@@ -0,0 +1,92 @@
...
+*inside* the terminal multiplexer must be set to `screen` or any
+compatible variant.
+
+You can check with `infocmp` that you have the proper entry for your
+terminal type in your terminfo database.
+If you want to use an alternate terminfo database, you can use
+`export TERMINFO=/path/to/terminfo`.
+
+Some OS (e.g. OS X 10.11) are shipping with an incomplete terminfo
+description for screen resulting, for example, in diff-highlight option
```
krobelus, line 43:
```suggestion
Some OS (e.g. OS X 10.11) are shipping with an incomplete terminfo
description for screen resulting, for example, in the diff-highlight option
```

PR #1027, opened 2020-07-23
Mention again that 'generic' keymap can be overridden

wsakernel:
There is a paragraph above to explain it in detail. However, redundancy
is not a bad thing in documentation IMO. In other words, I missed it
before.

Signed-off-by: Wolfram Sang <wsa@kernel.org>


wsakernel:
Also contained in PR #1078. Sorry, I forgot that I already sent this PR.

PR #1039, opened 2020-09-25
Some minor updates for scrolling and for a double-click in diff

mckellyln:
@jonas Thank you for awesome tig.  I use it every day.
Here are a few small updates that I think improve scrolling, and to not Enter when double-clicking on a diff line.
What do you think ?
thx,
-m

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -101,6 +101,8 @@ void
 scroll_view(struct view *view, enum request request)
 {
 	int lines = 1;
+	unsigned long orig_offset, orig_lineno;
+	unsigned long new_offset, new_lineno;
 
 	assert(view_is_displayed(view));
 
@@ -138,8 +140,14 @@ scroll_view(struct view *view, enum request request)
 		if (view->pos.offset + lines > view->lines)
 			lines = view->lines - view->pos.offset;
 
+		// do not scroll past bottom ...
+		if ((view->pos.offset + view->height + lines) > view->lines)
+			lines = view->lines - view->pos.offset - view->height;
+
 		if (lines == 0 || view->pos.offset + view->height >= view->lines) {
 			report("Cannot scroll beyond the last line");
+			// if wanted to move down when can no longer scroll ...
+			// move_view(view, REQ_MOVE_DOWN);
 			return;
 		}
 		break;
@@ -154,6 +162,8 @@ scroll_view(struct view *view, enum request request)
 
 		if (lines == 0) {
 			report("Cannot scroll beyond the first line");
+			// if wanted to move up when can no longer scroll ...
+			// move_view(view, REQ_MOVE_UP);
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


mckellyln:
Perhaps the 3 test diffs are due to the small changes in scrolling ?

krobelus:
I wanted to try this because I'm not sure what exactly changed.
Does it need mouse support? I tried `set mouse = true`, using [reasonable ncurses options](https://github.com/archlinux/svntogit-packages/blob/packages/ncurses/trunk/PKGBUILD#L25-L29), but the mouse behavior doesn't change at all, clicking just selects text, just like before I enabled  `mouse`.

mckellyln:
The changes are small but I found quite useful.  The changes are for both mouse or key scrolling.
1). double-click mouse in diff window no longer causes cursor to move down one line.
2). scrolling with mouse does not allow for scrolling past the last line at the bottom.
3). scrolling now moves the cursor same number of lines as I think is typical with other editors/viewers.
And I use these settings in my .tigrc for Shift-Down/Up for key-based scrolling:
```
bind generic   <ScrollFwd>  scroll-line-down
bind generic   <ScrollBack> scroll-line-up
# and for mouse -
set mouse              = yes
set mouse-scroll       = 10
set mouse-wheel-cursor = no
```

mckellyln:
I configured --with-ncursesw and have ncurses 6.2 installed.

src/view.c
```diff
@@ -164,7 +172,24 @@ scroll_view(struct view *view, enum request request)
 		die("request %d not handled in switch", request);
 	}
 
+	unsigned long orig_offset = view->pos.offset;
```
krobelus, line 175: I think the style convention here is to always declare variables at the beginning of a block (pre-C99).

src/view.c
```diff
@@ -164,7 +172,24 @@ scroll_view(struct view *view, enum request request)
...
 	do_scroll_view(view, lines);
+
+	unsigned long new_offset = view->pos.offset;
+	unsigned long new_lineno = view->pos.lineno;
+
+	int move_lines = (new_offset - orig_offset) - (new_lineno - orig_lineno);
+
+	int i;
+	if (move_lines < 0) {
+		for(i=0; i<-move_lines; i++)
```
krobelus, line 187: at first I thought this was a reverse dereference operator :laughing:

krobelus (commented):
Thanks, it works with this config! 1. and 2. are obvious improvements!
I somehow can't figure how to reproduce the difference from 3. :thinking:

mckellyln:
ok, I will move the declaration up, outside the block in a amended commit in a few.
For #3 move down a few lines so you are not at the top and then scroll a few lines.  Without my change the curosr moved with the scroll.  I wanted the cursor to remain at the same screen position.

mckellyln:
Amended commit.

krobelus:
> For item 3 move down a few lines so you are not at the top and then scroll a few lines

Right, I see it now, it's a really nice improvement as well. I had already noticed the difference before, but then re-enabled `mouse-wheel-cursor` which always moves the cursor, so it's not visible with that.

mckellyln:
Closing for now, may resubmit PR soon.

krobelus:
This still applies cleanly and works well, I don't think this needs more work (besides maybe small style changes).

mckellyln:
Resubmitted from another repo at: https://github.com/jonas/tig/pull/1060

PR #1042, opened 2020-10-12
docs: fix simple typo, fale -> false

timgates42:
There is a small typo in src/options.c.

Should read `false` rather than `fale`.


jonas:
Thank you!

PR #1047, opened 2020-11-03
Zsh completion fixes

felipec:
There's a bunch of issues regarding the zsh completion:

* The Zsh bash completion emulation was loaded at the wrong time
* The Git bash completion helpers were not used correctly
* The Zsh bash completion emulation can't be relied upon

These changes should fix the issues mentioned in pull #960 and #940.

### DIFF
--- a/contrib/tig-completion.bash
+++ b/contrib/tig-completion.bash
@@ -86,13 +86,6 @@ _tig() {
 	esac
 }
 
-# Detect if current shell is ZSH, and if so, load this file in bash
-# compatibility mode.
-if [ -n "$ZSH_VERSION" ]; then
-	autoload bashcompinit
-	bashcompinit
-fi
-
 # we use internal git-completion functions, so wrap _tig for all necessary
 # variables (like cword and prev) to be defined
 __git_complete tig _tig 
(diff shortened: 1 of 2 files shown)


llua:
The intention e073f1b seems to be to use the completer for `tig` provided by zsh, if that is the case, then `contrib/tig-completion.zsh` doesn't need to exist since zsh 4.3.12 (2011).

felipec:
> The intention e073f1b seems to be to use the completer for tig provided by zsh

No, not the one provided by Zsh, the one provided by Git:

https://git.kernel.org/pub/scm/git/git.git/tree/contrib/completion/git-completion.zsh

The one provided by Zsh would not load the tig bash completion function `__tig_main`, the one I wrote would.

contrib/tig-completion.bash
```diff
@@ -33,8 +33,6 @@ __tig_options="
 	-v --version
 	-h --help
 	-C
-	--
```
rohieb, line 36: Hmm, why did you reve these two lines? They are valid completion options :)

felipec: Because they confuse `__gitcomp`. Well, at least the `--` one.

Most commands do have the `--` option, but none that I'm aware of actually add it to the completion.

rohieb: okay, then that sounds fine to me too.

koutcher (approved):
A big thanks for looking into this issue (especially as my previous attempt in 2.5.1 did more bad than good).

PR #1048, opened 2020-11-09
Add ncurses extended key support

mckellyln:
Keys added, if terminfo reports feature -
``` 
    { "kUP3",   "Alt-Up",              "A-Up"          },
    { "kUP4",   "Alt-Shift-Up",        "A-S-Up"        },
    { "kUP5",   "Ctrl-Up",             "C-Up"          },
    { "kUP6",   "Ctrl-Shift-Up",       "C-S-Up"        },
    { "kUP7",   "Alt-Ctrl-Up",         "A-C-Up"        },
    { "kDN3",   "Alt-Down",            "A-Down"        },
    { "kDN4",   "Alt-Shift-Down",      "A-S-Down"      },
    { "kDN5",   "Ctrl-Down",           "C-Down"        },
    { "kDN6",   "Ctrl-Shift-Down" ,    "C-S-Down"      },
    { "kDN7",   "Alt-Ctrl-Down",       "A-C-Down"      },
    { "kRIT3",  "Alt-Right",           "A-Right"       },
    { "kRIT4",  "Alt-Shift-Right",     "A-S-Right"     },
    { "kRIT5",  "Ctrl-Right",          "C-Right"       },
    { "kRIT6",  "Ctrl-Shift-Right",    "C-S-Right"     },
    { "kRIT7",  "Alt-Ctrl-Right",      "A-C-Right"     },
    { "kLFT3",  "Alt-Left",            "A-Left"        },
    { "kLFT4",  "Alt-Shift-Left",      "A-S-Left"      },
    { "kLFT5",  "Ctrl-Left",           "C-Left"        },
    { "kLFT6",  "Ctrl-Shift-Left",     "C-S-Left"      },
    { "kLFT7",  "Alt-Ctrl-Left",       "A-C-Left"      },
    { "kIC3",   "Alt-Insert",          "A-Insert"      },
    { "kIC4",   "Alt-Shift-Insert",    "A-S-Insert"    },
    { "kIC5",   "Ctrl-Insert",         "C-Insert"      },
    { "kIC6",   "Ctrl-Shift-Insert",   "C-S-Insert"    },
    { "kIC7",   "Alt-Ctrl-Insert",     "A-C-Insert"    },
    { "kDC3",   "Alt-Delete",          "A-Delete"      },
    { "kDC4",   "Alt-Shift-Delete",    "A-S-Delete"    },
    { "kDC5",   "Ctrl-Delete",         "C-Delete"      },
[...]

PR #1049, opened 2020-11-13
Disable ncurses extended key values so that esc-codes can be used in …

mckellygit:
…mappings for Shift/Ctrl/Alt Up/Dn/Lt/Rt/Hm/En/In/De/PgUp/PgDn

This is the other way to go, as compared to PR https://github.com/jonas/tig/pull/1048

thx,
-m

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -708,6 +708,17 @@ init_display(void)
 		use_scroll_redrawwin = true;
 		use_scroll_status_wclear = false;
 	}
+
+#if defined(NCURSES_EXT_FUNCS)
+    // Disable extended keys so that esc-codes will be received
+    // instead of extended key values (> KEY_MAX).
+    // Then these keys can be mapped in .tigrc etc.
+    int i;
+    for (i=KEY_MAX; i<2000; i++) {
+        keyok(i, false);
+    }
+#endif
+
 }
 
 static bool


koutcher:
Merged manually in aee8eb1.

mckellygit:
@koutcher thx.  Something is weird tho, in my original PR a binding such as:
```
bind generic   <Esc>[1;5B   scroll-line-down
```
worked fine - but now it does not, and I need to use:
```
bind generic   <Ctrl-[>[1;5B   scroll-line-down
```
I will dig into it and post what I find.

mckellygit:
@koutcher its a problem with my other commit that was merged and the values for Esc and C-_
I'll post a PR to fix it asap.

mckellygit:
ok, PR added: https://github.com/jonas/tig/pull/1058

koutcher:
Good catch. I should have spotted that during the review. I was so focussed on the possible side-effect of changing the < in <= that I was happy enough when I checked that CTRL('z') was handled separately and I completely overlooked that when going from 25 to 30, 27 was also invited to the party. Hopefully 28 is only ^\.

mckellygit:
@koutcher @jonas just an fyi - I was working with xaizek on vifm for this same issue and he suggested a cleaner way to solve this.
We added this code -
```
#if defined(NCURSES_EXT_FUNCS) && NCURSES_EXT_FUNCS >= 20081102
	/* Disable 'unsupported' extended keys so that esc-codes will be
	 * received instead of unnamed extended keycode values (> KEY_MAX).
	 * These keys can then be mapped with esc-codes in rc files.
	 * An example could be <c-down>:
	 *     bind generic <esc>[1;5B scroll-line-down
	 * as without this change <c-down> can return a keycode value of 531
	 * (terminfo name kDN5), which is larger than KEY_MAX and has no
	 * pre-defined curses key name.
	 * NOTE: this MUST be called before initscr() */
	use_extended_names(FALSE);
#endif
```
just BEFORE initscr() is called.

I just thought I'd mention it.  If you want I can create a PR to remove the keyok() loop and use this instead ?
thx,
-m

koutcher:
It is much cleaner indeed. Why 20081102 ? `NCURSES_EXT_FUNCS` macro was only made visible in ncurses v5_6_20071117 while `use_extended_names` appeared in v4_2_990301, so using `#ifdef NCURSES_VERSION` instead would allow the last CentOS 5 users to also benefit from the PR.

mckellygit:
Ok sure, NCURSES_VERSION is good - I just used an old version thinking it was old enough.
I should have copied the same macro name you used before.
I'll create a PR

mckellygit:
Created https://github.com/jonas/tig/pull/1074

PR #1055, opened 2020-12-02
Handle undefined __git_complete

darcyparker:
Addresses https://github.com/jonas/tig/issues/1011

As identified in https://github.com/jonas/tig/issues/1011#issuecomment-736922515, there are cases where `__git_complete` is undefined because it is loaded dynamically by bash_completions.

As suggested in https://github.com/jonas/tig/issues/1011#issuecomment-736938541, using `pkg-config` to find `git-completion` script, which defines `__git_complete`.  And using approach similar to https://github.com/felipec/git-completion/blob/master/git-completion.zsh#L28-L49

### DIFF
--- a/contrib/tig-completion.bash
+++ b/contrib/tig-completion.bash
@@ -1,5 +1,5 @@
 # bash/zsh completion for tig
-# 
+#
 # Copyright (C) 2019 Roland Hieber, Pengutronix
 # Copyright (C) 2007-2010 Jonas fonseca
 #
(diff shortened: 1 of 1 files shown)


contrib/tig-completion.bash
```diff
@@ -29,6 +29,27 @@
 #       at source time then all lookups will be done on demand,
 #       which may be slightly slower.
 
+#tig-completion requires __git_complete
+#* If not defined, source git completions script so __git_complete is available
+if ! declare -f __git_complete &>/dev/null; then
+  _bash_completion=$(pkg-config --variable=completionsdir bash-completion 2>/dev/null) ||
```
rohieb, line 35: Nitpick: the indentation should be consistent with the rest of the file.

darcyparker: Thank you. I will fix that.

rohieb:
Yep, LGTM now.

koutcher:
Thanks.

PR #1056, opened 2020-12-11
Use id column for commit id only

### LINKED ISSUE #1025: field 'id' in reflog-view does not show SHA hash
Setting

```
set reflog-view = id:width=40
```

in `.tigrc` does not result in the view showing the SHA hash. Instead it still only prints the symbolic refs like `HEAD@{...}`. I would really like to see the SHAs directly as well, because this is how default git-reflog behaves. So ideally I should be able to do

```
set reflog-view = id:width=40 id:yes
```

to see the SHA and then also the `HEAD@{...}` ref.

- Version: 2.5.1
- Platform: Mac

koutcher:
That way, you can better figure out identical references when browsing
the reflog view. Show the reflog name in the title window instead. In
the stash view, the id column also shows the commit id instead of the
reference.

Closes #1025

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -364,8 +364,6 @@ main_get_column_data(struct view *view, const struct line *line, struct view_col
 	column_data->author = commit->author;
 	column_data->date = &commit->time;
 	column_data->id = commit->id;
-	if (state->reflogs)
-		column_data->reflog = state->reflog[line->lineno - 1];
 
 	column_data->commit_title = commit->title;
 	if (state->with_graph) {
@@ -587,9 +585,16 @@ main_select(struct view *view, struct line *line)
 		string_ncopy(view->ref, commit->title, strlen(commit->title));
 		status_stage_info(view->env->status, line->type, NULL);
 	} else {
+		struct main_state *state = view->private;
 		const struct ref *ref = main_get_commit_refs(line, commit);
 
-		string_copy_rev(view->ref, commit->id);
+		if (state->reflogs) {
+			assert(state->reflogs >= line->lineno);
+			string_ncopy(view->ref, state->reflog[line->lineno - 1],
+				     strlen(state->reflog[line->lineno - 1]));
+		} else {
+			string_copy_rev(view->ref, commit->id);
+		}
 		if (ref)
 			ref_update_env(view->env, ref, true);
 	}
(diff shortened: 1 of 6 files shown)


krobelus:
I wonder if it makes sense to color-code the commit ids like in the blame view, so it's even easier to spot identical refereences. Probably a bit too noisy though, since most references are different.

koutcher:
From the test suite point of view, this PR doesn't break anything... @jonas, I tried to track the origin of this behaviour (i.e. using the id column to display reflog) and went as far as e32cad8 but I can't figure out why it was introduced. Maybe you can remember. It looks it was related to the stash view. Was it a way to show the stash reference when it was not displayed in the title window ?

koutcher:
> I wonder if it makes sense to color-code the commit ids like in the blame view, so it's even easier to spot identical refereences. Probably a bit too noisy though, since most references are different.

We just need a small change in tigrc: `id:yes,color`. As it is hash based and the palette is quite small it is not bullet proof but it helps already.

jonas:
Yes, the ID colouring was mainly for the blame view and as you say very limited and confusing to use it outside for other views due to collisions. It might make sense to be able to set a specific colour where `color=by-id` could be the current behavior.

jonas:
> @jonas, I tried to track the origin of this behaviour (i.e. using the id column to display reflog) and went as far as [e32cad8](https://github.com/jonas/tig/commit/e32cad84bc199ec3113fa9448082eb8b9b9b080f) but I can't figure out why it was introduced. Maybe you can remember. It looks it was related to the stash view. Was it a way to show the stash reference when it was not displayed in the title window ?

Yes, from what I remember, the idea was to show the `@stash{...}` but the ID column was later hidden?!? I agree that using the ID column for SHAs only makes sense. A future change could change the reflog view to show the "symbolic name" in the title bar, similar to what we do in the stash view.

src/main.c
```diff
@@ -587,9 +585,16 @@ main_select(struct view *view, struct line *line)
...
 		status_stage_info(view->env->status, line->type, NULL);
 	} else {
+		struct main_state *state = view->private;
 		const struct ref *ref = main_get_commit_refs(line, commit);
 
-		string_copy_rev(view->ref, commit->id);
+		if (state->reflogs) {
+			assert(state->reflogs >= line->lineno);
+			string_ncopy(view->ref, state->reflog[line->lineno - 1],
+				     strlen(state->reflog[line->lineno - 1]));
```
jonas, line 594: :+1: This will show the reflog ID in the title bar.

jonas (approved):

koutcher:
Thanks @jonas.

PR #1060, opened 2020-12-29
Some minor updates for scrolling and for a double-click in diff

mckellygit:
Original PR was in: https://github.com/jonas/tig/pull/1039
I just moved my repo and re-submitted PR.
This change basically keeps cursorline at same position when scrolling, which I think behaves more similar to most other editors.


krobelus:
The failing tests probably just need fixture changes.

mckellygit:
@koutcher looks good.
I tested it some and it works well.

PR #1064, opened 2020-12-31
Allow to generate a compilation database

krobelus:
This is useful for some Clang-based tools, in particular language servers like clangd or ccls.

I mostly took this mostly from Git's Makefile.  The awkward if-statement
is because the "missing_compdb_dir" trick that Git uses didn't work for
some reason.

Keep in mind that not all compilers support the -MJ option. Clang does, so this
command should give you a fresh compile_commands.json

	make CC=clang GENERATE_COMPILATION_DATABASE=yes clean all-debug

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,12 +85,51 @@ else
 TAR ?= tar
 endif
 
+GENERATE_COMPILATION_DATABASE ?= no
+ifeq ($(GENERATE_COMPILATION_DATABASE),yes)
+compdb_check = $(shell $(CC) $(ALL_CFLAGS) \
+	-c -MJ /dev/null \
+	-x c /dev/null -o /dev/null 2>&1; \
+	echo $$?)
+ifneq ($(compdb_check),0)
+override GENERATE_COMPILATION_DATABASE = no
+$(warning GENERATE_COMPILATION_DATABASE is set to "yes", but your compiler does not \
+support generating compilation database entries)
+endif
+else
+ifneq ($(GENERATE_COMPILATION_DATABASE),no)
+$(error please set GENERATE_COMPILATION_DATABASE to "yes" or "no" \
+(not "$(GENERATE_COMPILATION_DATABASE)"))
+endif
+endif
+
+compdb_dir = compile_commands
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -338,7 +339,8 @@ DEPS_CFLAGS ?= -MMD -MP -MF .deps/$*.d
 
 %.o: %.c $(CONFIG_H)
 	@mkdir -p .deps/$(*D)
-	$(QUIET_CC)$(CC) -I. -Iinclude $(CFLAGS) $(DEPS_CFLAGS) $(CPPFLAGS) -c -o $@ $<
+	@if [ x$(GENERATE_COMPILATION_DATABASE) = xyes ]; then mkdir -p $(compdb_dir); fi
```
krobelus, line 342: Git's Makefile uses

```make
missing_compdb_dir = $(compdb_dir)
$(missing_compdb_dir):
	@mkdir -p $@
```

and then builds objects like

```Makefile
$(C_OBJ): %.o: %.c GIT-CFLAGS $(missing_dep_dirs) $(missing_compdb_dir)
	$(QUIET_CC)$(CC) -o $*.o -c $(dep_args) $(compdb_args) $(ALL_CFLAGS) $(EXTRA_CPPFLAGS) $<
```

however, adding ` $(missing_compdb_dir)` to line 339 didn't work, maybe someone can spot why.

Makefile
```diff
@@ -406,6 +408,41 @@ test/API.adoc: test/tools/libtest.sh
...
+endif
+
+compdb_dir = compile_commands
+ifeq ($(GENERATE_COMPILATION_DATABASE),yes)
+compdb_file = $(compdb_dir)/$(subst /,-,$@.json)
+compdb_args = -MJ $(compdb_file)
+else
+compdb_args =
+endif
+
```
koutcher, line 437: Move this part at line 88 to make `$(missing_compdb_dir)` work.

krobelus: Yep, thanks! I'm not sure why because I don't see where the order would matter but that's okay :shrug:

PR #1069, opened 2021-01-06
Use comparison instead of destructive assignment

krobelus:
I found this when working on the CI, running, see
https://builds.sr.ht/~krobelus/job/388884 which runs

	CC=clang TIG_BUILD=config.make tools/travis.sh

The warning from clang was:

	src/diff.c:377:27: error: converting the enum constant to a boolean [-Werror,-Wint-in-bool-context]
	                   (type = LINE_DEFAULT ||

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -373,10 +373,7 @@ diff_common_read(struct view *view, const char *data, struct diff_state *state)
 
 	} else if (opt_word_diff && state->reading_diff_chunk &&
 		   /* combined diff format is not using word diff */
-		   !state->combined_diff &&
-		   (type = LINE_DEFAULT ||
-		    /* ADD and DEL are only valid in regular diff hunks */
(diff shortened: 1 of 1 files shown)


src/diff.c
```diff
@@ -374,7 +374,7 @@ diff_common_read(struct view *view, const char *data, struct diff_state *state)
 	} else if (opt_word_diff && state->reading_diff_chunk &&
 		   /* combined diff format is not using word diff */
 		   !state->combined_diff &&
-		   (type = LINE_DEFAULT ||
+		   (type == LINE_DEFAULT ||
 		    /* ADD and DEL are only valid in regular diff hunks */
 		    type == LINE_DIFF_ADD || type == LINE_DIFF_DEL)) {
```
koutcher, line 379: Good catch! Actually, the 3 tests on type are redundant and could be removed.

Issue #858, opened 2018-07-27
Compile error


dscho:
New in v2.4.1 (was not there in v2.4.0):

```
2018-07-27T10:36:13.4602087Z src/graph-v2.c: In function 'colors_init':
2018-07-27T10:36:13.4604561Z src/graph-v2.c:217:3: error: unknown type name 'uint'; did you mean 'int'?
2018-07-27T10:36:13.4606435Z    uint size = 500;
2018-07-27T10:36:13.4607091Z    ^~~~
2018-07-27T10:36:13.4608026Z    int
```

Let me know if you need me to open a PR, or whether it is quicker for you to just fix it.

chros73:
Hm, I just compiled v2.4.1 on CentOS 6.10 (gcc 4.4.7) and on Ubuntu 14.04 (gcc 4.8.4) without any issue (using this [build script](https://github.com/chros73/build-tig)).
What OS, distrib do you use?

jonas:
The only change between the two versions is https://github.com/jonas/tig/commit/62d559a382af15bec3fe6fa5d02cbdb1999a5b6d
Strange

dscho:
Hmpf.

@chros73 I am using the Git for Windows distribution 😛 

I don't have time right now, especially not to hunt down a bug that I only encountered today when I tried to upgrade the `tig` included in Git for Windows, but I will try to find some time (and not to forget about it).

koutcher:
The uint should be replaced with size_t anyway...

PR #1070, opened 2021-01-07
Switch from Travis CI to GitHub Actions

### LINKED ISSUE #1061: Consider moving away from Travis
Travs has gone [freemium](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing). Open source projects can ask for extra build credits but [those are limited](https://news.ycombinator.com/item?id=25338983).

As a result Travis does not run on recent PRs. We could move to GitHub Actions or similar. HTH

krobelus:
Closes #1061

Simplistic port of the old Travis configuration. The results looks like this: [ubuntu.yml](https://github.com/krobelus/tig/actions/runs/468621862), [macos.yml](https://github.com/krobelus/tig/actions/runs/468621859).

This runs all eight combinations of compilers and build flags. We could select an interesting subset. Additionally we could run tests on other OS.


koutcher:
This is really great! I think we should move Cygwin builds from AppVeyor to GitHub Actions as well.
The full matrix could be a bit overkill (and time consuming). Previously address-sanitizer was only run with clang and valgrind with gcc.
Badges will have to be updated in README.adoc.

krobelus:
Ok I've updated the badges, and made the matrix more selective, so only 6 configurations are run instead of 8 on ubuntu.
I also added a Windows build but it fails to find <sys/wait.h>, which should be included in the Cygwin package [cygwin-devel](https://cygwin.com/cgi-bin2/package-cat.cgi?file=x86_64%2Fcygwin-devel%2Fcygwin-devel-3.1.7-1&grep=sys%2Fwait).
I didn't set up cygwin, so debugging is somewhat difficult. I don't think GitHub Actions provide SSH access.

README.adoc
```diff
@@ -2,8 +2,9 @@ Tig: text-mode interface for Git
 ================================
 :docext: adoc
 
-image:https://secure.travis-ci.org/jonas/tig.svg?branch=master[Travis Build,link=https://travis-ci.org/jonas/tig]
-image:https://ci.appveyor.com/api/projects/status/jxt1uf52o7r0a8r7/branch/master?svg=true[AppVeyor Build,link=https://ci.appveyor.com/project/fonseca/tig]
+image:https://github.com/jonas/tig/workflows/Linux/badge.svg[Linux CI,link=https://github.com/jonas/tig/actions?query=workflow%3A.github%2Fworkflows%2Flinux.yml]
+image:https://github.com/jonas/tig/workflows/MacOS/badge.svg[MacOS CI,link=https://github.com/jonas/tig/actions?query=workflow%3A.github%2Fworkflows%2Fmacos.yml]
+image:https://github.com/jonas/tig/workflows/Windows/badge.svg[Ubuntu CI,link=https://github.com/jonas/tig/actions?query=workflow%3A.github%2Fworkflows%2Fwindows.yml]
```
koutcher, line 7: Typo: Ubuntu -> Windows. I believe Mac OS should also now be referred as macOS.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,25 @@
...
+          git config --global core.autocrlf false
+          git config --global core.eol lf
+      - uses: actions/checkout@v2
+      - name: Set up Cygwin
+        uses: egor-tensin/setup-cygwin@v2
+        with:
+          packages: git gcc-code make automake libiconv-devel libncurses-devel cygwin-devel glibc
+      - name: Test Tig
+        env:
+          CC: gcc
```
koutcher, line 20: Maybe setting the PATH environment variable could make a difference as appveyor.yml has a line with ` - cmd: SET PATH=C:\%TARGET%\bin;%PATH%`.

krobelus: Right, we actually need to prepend to the path to use the right binaries. I have made this work on a Windows system now. The tests work fine (but are really slow). I don't know a workaround for https://github.com/actions/runner/issues/241 that works on Windows, so I commented out `make test`. We can probably just leave it like that for now.

koutcher: Nice job. As we run the full test on Linux, it seems to be enough to just test the build on macOS and Windows. We can wait a feedback from @jonas to see if appveyor.yml can go as well.

krobelus: The Windows job is not very interesting because with Cygwin it's using the same GNU toolchain as Linux.
It is, however useful as documentation of how to quickly get Tig up and running on Windows. I'm no longer using the GitHub Actions library that abstract installing cygwin, because a single `choco install cygwin` is enough; that makes it easier to recreate locally as well.

Maybe I'll try to run tests on macOS as well (provided the `script` workaround to get a TTY works there).

Could also be interesting to test on different libcs and other LInux distribution (which would also document which packages to install). But that's all not super important I guess.

koutcher: We have seen in the past some Windows specific compilation issues (#858), so I think it is worth having the Windows job.
`script` should work on macOS but I don't know how we can run the tests with the Brew build.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,17 @@
...
+  install:
+    runs-on: macos-latest
+
+    strategy:
+      matrix:
+        compiler: [clang, gcc]
+
+    steps:
+      - name: Install latest Tig
+        run: CC=${{ matrix.compiler }} HOMEBREW_NO_AUTO_UPDATE=1 brew install --HEAD tig
```
koutcher, line 17: I don't think that Homebrew will honour the CC variable here. `HOMEBREW_CC=${{ matrix.compiler }}` may be a better choice.

krobelus: makes sense, done

.github/workflows/windows.yml
koutcher: Adding this step to macos.yml should make Homebrew run the test suite for us:
```
    - name: Patch
      run: |
        cd "$(brew --repo homebrew/core)" && patch -p1 <<\EOF
diff --git a/Formula/tig.rb b/Formula/tig.rb
index 2e8c125660..f76d6bca1c 100644
--- a/Formula/tig.rb
+++ b/Formula/tig.rb
@@ -26,9 +26,11 @@ class Tig < Formula
   depends_on "readline"
 
   def install
+    ENV.deparallelize
     system "./autogen.sh" if build.head?
     system "./configure", "--prefix=#{prefix}", "--sysconfdir=#{etc}"
     system "make"
+    system "make", "test"
     # Ensure the configured `sysconfdir` is used during runtime by
     # installing in a separate step.
     system "make", "install", "sysconfdir=#{pkgshare}/examples"
EOF
```

krobelus: yep, that works

.github/workflows/windows.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: Install dependencies
+        run: choco install -y --no-progress --source=cygwin git gcc-core make automake pkg-config libiconv-devel libncurses-devel
+      - name: Test Tig
+        run: |
+          $Env:PKG_CONFIG_PATH = "C:\tools\cygwin\lib\pkgconfig"
+          $Env:PATH = "C:\tools\cygwin\bin;" + $Env:PATH
+          make configure
+          sh configure
+          make
+          # make test # TODO There is no TTY on Windows, see https://github.com/actions/runner/issues/241
```
koutcher, line 25: Have you tried running `make test` on Windows ? When running in CMD.EXE, there is no TTY anyway, so the script hack may not be necessary for the Cygwin build...

krobelus: yeah, it [fails](https://github.com/krobelus/tig/runs/1676952058?check_suite_focus=true) due to the missing TTY

On a real Windows machine the tests work (but run extremely slowly).

koutcher: Ok then I believe we should drop windows.yml for now and keep relying on the AppVeyor build for Windows CI until there is a solution for this problem. Don't forget to put the AppVeyor badge back in README.adoc and I think we are good to go.

krobelus: ok, done

PR #1074, opened 2021-01-23
Disable ncurses extended keycode values.  A better method than prev PR #1049

mckellygit:
Hi,
Here is a cleaner fix than previous PR https://github.com/jonas/tig/pull/1049
Thanks to xaizek also.
thx,
-m

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -648,6 +648,19 @@ init_display(void)
 	if (atexit(done_display))
 		die("Failed to register done_display");
 
+#ifdef NCURSES_VERSION
+	/* Disable 'unsupported' extended keys so that esc-codes will be
+	 * received instead of unnamed extended keycode values (> KEY_MAX).
+	 * These keys can then be mapped with esc-codes in .tigrc.
+	 * An example would be <c-down>:
+	 *     bind  generic  <esc>[1;5B  scroll-line-down
+	 * Without this change <c-down> could return a keycode value of 531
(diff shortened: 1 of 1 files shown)


koutcher:
Have you tried mouse support after this PR ? Looks broken to me.

Btw, I'm also looking at your other PR #1060 (that's how I found out as I'm not using the mouse support otherwise), I have some changes to propose after the review, I you agree I can push them to your PR branch for you to check them.

mckellygit:
Hi,
I just tested this PR with:
```
set mouse = yes
set mouse-wheel-cursor = no (and yes)
```
And mouse clicks seem to be ok for me.
I'll keep testing and try some different terminals.
Sure, glad to review/help out with the PR # 1060 :-)

koutcher:
I double checked and can confirm that on macOS with Terminal, using
```
$ tig -v
tig version 2.5.1-48-g0bb11a2
ncursesw version 6.2.20200212
readline version 8.1
```
the mouse support is definitely broken while everything goes back to normal when recompiling master.
I just uncommented the #set mouse = yes in the default tigrc.

mckellygit:
Ok.  Lets just leave it as is then, with the keyok(false) method and I'll close this PR.
I'll research it more some time in the future.

koutcher:
The problem occurs with `xterm` (and derived) terminal descriptions. It seems to be linked to changes introduced in ncurses 6.1 regarding kmous capability.

mckellygit:
ok, I'll close this PR now.

PR #1079, opened 2021-01-28
Allow to change directory from the next argument

deeagle001:
The -C change directory option only accepts dir within the same argv argument. The git tool allows this value to be in the next argument. This PR allows providing the change directory as the next argument after -C.

### DIFF
--- a/src/tig.c
+++ b/src/tig.c
@@ -488,8 +488,10 @@ parse_options(int argc, const char *argv[], bool pager_mode)
 	for (i = 1; i < argc; i++) {
 		const char *opt = argv[i];
 		if (!strncmp(opt, "-C", 2)) {
-			if (chdir(opt + 2))
-				die("Failed to change directory to %s", opt + 2);
+			const char *dir = strlen(opt) == 2 && i + 1 < argc ? \
+					  argv[++i] : opt + 2;
+			if (chdir(dir))
+				die("Failed to change directory to %s", dir);
 			continue;
 		} else {
 			break;


koutcher:
To be honest, if I had to name one contribution I wished I didn't do it would probably be `-C`... It already required #911 to fix some regressions and I realize that I need `-C` (detect renames) more often than `-C/path/to/repo`. Actually, Git itself only accepts `-C /path/to/repo` and not `-C/path/to/repo`, so I would propose to go even further with: 
```patch
diff --git a/src/tig.c b/src/tig.c
index ed4388f..f02566a 100644
--- a/src/tig.c
+++ b/src/tig.c
@@ -389,7 +389,7 @@ static const char usage_string[] =
 "  +<number>       Select line <number> in the first view\n"
 "  -v, --version   Show version and exit\n"
 "  -h, --help      Show help message and exit\n"
-"  -C<path>        Start in <path>";
+"  -C <path>       Start in <path>";
 
 void
 usage(const char *message)
@@ -489,9 +489,10 @@ parse_options(int argc, const char *argv[], bool pager_mode)
 	/* Options that must come before any subcommand. */
 	for (i = 1; i < argc; i++) {
 		const char *opt = argv[i];
-		if (!strncmp(opt, "-C", 2)) {
-			if (chdir(opt + 2))
-				die("Failed to change directory to %s", opt + 2);
+		if (!strcmp(opt, "-C") && i + 1 < argc && *argv[i + 1] != '-') {
+			const char *dir = argv[++i];
+			if (chdir(dir))
+				die("Failed to change directory to %s", dir);
 			continue;
 		} else {
 			break;
```
That would allow the `-C /path/to/repo` syntax and, in case `-C` is used alone, it would be passed to the second stage of command line parsing where it would be added to diff options. What do you think ?

krobelus:
Yeah, the earlier we change this, the better. With `-C foo`, shell compeltion will work OOTB.
Since Git doesn't allow `-Cfoo`, it feels okay to mimic that. It is breaking but avoids confusion in the long term.

PR #1081, opened 2021-02-10
Expose the line number before a diff as %(lineno_old)

krobelus:
I use Tig to write review comments on GitLab (hopefully soon GitHub as well).
To comment on specific lines of a diff, the GitLab API requires both the
line number _before_ the diff, and _after_ the diff.  Additionally, it needs
the line type (added, deleted / context line). This patch exposes %(lineno_old)
which is the dual to %(linenno). It could also be useful when viewing
"reverse" diffs.

### DIFF
--- a/doc/manual.adoc
+++ b/doc/manual.adoc
@@ -175,6 +175,8 @@ following variables.
 			 "." if undefined.
 |%(file)		|The currently selected file.
 |%(lineno)		|The currently selected line number. Defaults to 0.
+|%(lineno_old)		|The currently selected line number, before the diff
+			 was applied. Defaults to 0.
 |%(ref)			|The reference given to blame or HEAD if undefined.
 |%(revargs)		|The revision arguments passed on the command line.
 |%(fileargs)		|The file arguments passed on the command line.
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -11,6 +11,7 @@ Improvements:
  - Add line-graphics = auto. (#834)
  - Allow maxwidth to be expressed in % of the view width.
  - Pass stash args through. (#1022)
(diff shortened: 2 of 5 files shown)


doc/manual.adoc
```diff
@@ -175,6 +175,7 @@ following variables.
 			 "." if undefined.
 |%(file)		|The currently selected file.
 |%(lineno)		|The currently selected line number. Defaults to 0.
+|%(lineno_old)		|The selected line number in the pre-image of a diff. Defaults to 0.
```
koutcher, line 178: Just wondering if we can't find a better name for lineno_old. Not that easy though...

krobelus: Maybe `lineno_original`, that's what Wikipedia's diff page uses.

When browsing a patch that has not yet been applied to the working tree, this line number can be used to go to the exact line under the cursor. Of course this can only work if the cursor is on a (future) deleted, or context line. If the cursor is in a sequence of added lines, this line number will be on the first of the sequence.

I briefly considered making it symmetrical by adding an alias, so we can have something like `lineno_before` and `lineno_after` bur that's probably, not worth it.

koutcher: So be it for lineno_old, it not worth the time, you're right.

koutcher: `doc/tigrc.5.adoc` should be impacted as well.

krobelus: "old" seems alright because `git-diff` uses this terminology - it has options like `--output-indicator-old`.

`pre-image` was bogus though. It implies a function / unique mapping, which doesn't make sense here. I think I got that from the message that rerere prints ("Recorded preimage ...").

I've updated the docs and commit message

PR #1082, opened 2021-02-10
Update %(text) when navigating diff contents

krobelus:
Previously we'd only update %(text) in the header of the diff view.  Also
update %(text) when inside the diff contents. This allows to determine if
the current line is added, deleted or context, which allows to give precise
location information when talking to the GitLab API for example.

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -830,9 +830,9 @@ diff_common_select(struct view *view, struct line *line, const char *changes_msg
 			view->env->blob[0] = 0;
 		} else {
 			string_ncopy(view->ref, view->ops->id, strlen(view->ops->id));
-			pager_select(view, line);
 		}
 	}
+	pager_select(view, line);
 }
 
 static void


src/diff.c
```diff
@@ -830,8 +830,8 @@ diff_common_select(struct view *view, struct line *line, const char *changes_msg
 			view->env->blob[0] = 0;
 		} else {
 			string_ncopy(view->ref, view->ops->id, strlen(view->ops->id));
-			pager_select(view, line);
 		}
+		pager_select(view, line);
```
koutcher, line 834: And why not 2 levels up ?
```
                        string_ncopy(view->ref, view->ops->id, strlen(view->ops->id));
                }
        }
        pager_select(view, line);
}
```

koutcher: It looks more in line with the documentation (`The text of the currently selected line.`) and I don't see any drawback to do it for all lines.

krobelus: Yep, that seems more logical. No reason to leave out the stat-lines.

f1302c0780f834af1b8298029d1619118ab01388 added `%(text)` only in the pager view, and now we add it to the diff and stage views. There are still some other views that don't have it, but that's not a bad thing IMO.
For example, in the main view this seems mostly redundant because the user can compose the line with other substitutions already.

PR #1085, opened 2021-02-19
Format time_t values portably and fix other values too

### LINKED ISSUE #1084: Can't use %ld to portably print time_t values
Printing values of type `time_t` using the `%ld` format string assumes a specific `time_t` representation which may not hold. For example, on [musl libc](https://musl.libc.org) `time_t` is a 64 bit value, even on 32 bit systems thus `time_t` is not a `long int` on 32 bit musl architectures. See the [musl time64 Release Notes](https://musl.libc.org/time64.html) for more information. Currently, this causes the following compiler warning when attempting to compile tig on a 32 bit musl system:

```
src/util.c:163:48: warning: format '%ld' expects argument of type 'long int', but argument 6 has type 'time_t' {aka 'long long int'} [-Wformat=]
  163 |    if (!string_nformat(buf, buflen, NULL, "%s%ld%c",
      |                                              ~~^
      |                                                |
      |                                                long int
      |                                              %lld
  164 |         now.tv_sec >= timestamp ? "" : "-",
  165 |         seconds, reldate[i].compact_symbol))
      |         ~~~~~~~                                 
      |         |
      |         time_t {aka long long int}
src/util.c:168:52: warning: format '%ld' expects argument of type 'long int', but argument 5 has type 'time_t' {aka 'long long int'} [-Wformat=]
  168 |   } else if (!string_nformat(buf, buflen, NULL, "%ld %s%s %s",
      |                                                  ~~^
      |                                                    |
      |                                                    long int
      |                                                  %lld
  169 |         seconds, reldate[i].name,
      |         ~~~~~~~                                     
      |         |
      |         time_t {aka long long int}
```

It will likely result in a wrong output on these system as well, didn't test this though.

koutcher:
Fixes #1084

### DIFF
--- a/src/draw.c
+++ b/src/draw.c
@@ -322,7 +322,7 @@ draw_lineno_custom(struct view *view, struct view_column *column, unsigned int l
 		return false;
 
 	if (lineno == 1 || (lineno % interval) == 0) {
-		static char fmt[] = "%ld";
+		static char fmt[] = "%3u";
 
 		fmt[1] = '0' + digits3;
 		if (string_format(number, fmt, lineno))
--- a/src/ui.c
+++ b/src/ui.c
@@ -170,7 +170,7 @@ file_finder_draw(struct file_finder *finder)
 
 	wmove(finder->win, finder->height - 1, 0);
 	wbkgdset(finder->win, get_line_attr(NULL, LINE_TITLE_FOCUS));
-	wprintw(finder->win, "[finder] file %d of %d", pos->lineno + 1, finder->lines);
+	wprintw(finder->win, "[finder] file %lu of %zu", pos->lineno + 1, finder->lines);
 	wclrtoeol(finder->win);
 	wrefresh(finder->win);
 }
--- a/src/util.c
+++ b/src/util.c
@@ -146,13 +146,13 @@ get_relative_date(const struct time *time, char *buf, size_t buflen, bool compac
 {
 	struct timeval now;
 	time_t timestamp = time->sec + time->tz;
-	time_t seconds;
+	long long seconds;
 	int i;
 
 	if (time_now(&now, NULL))
 		return "";
 
-	seconds = now.tv_sec < timestamp ? timestamp - now.tv_sec : now.tv_sec - timestamp;
+	seconds = now.tv_sec < timestamp ? (long long) difftime(timestamp, now.tv_sec) : (long long) difftime(now.tv_sec, timestamp);
 
 	for (i = 0; i < ARRAY_SIZE(reldate); i++) {
 		if (seconds >= reldate[i].interval && reldate[i].interval)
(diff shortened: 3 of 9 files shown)


src/util.c
```diff
@@ -160,12 +160,12 @@ get_relative_date(const struct time *time, char *buf, size_t buflen, bool compac
...
 		seconds /= reldate[i].in_seconds;
 		if (compact) {
-			if (!string_nformat(buf, buflen, NULL, "%s%ld%c",
+			if (!string_nformat(buf, buflen, NULL, "%s%"PRIdMAX"%c",
 				    now.tv_sec >= timestamp ? "" : "-",
 				    seconds, reldate[i].compact_symbol))
 				return "";
 
-		} else if (!string_nformat(buf, buflen, NULL, "%ld %s%s %s",
+		} else if (!string_nformat(buf, buflen, NULL, "%"PRIdMAX" %s%s %s",
```
ericonr, line 168: If you use this format specifier, what you should do is cast the value as well (so `(intmax_t)value` iirc). Otherwise this will read 64 bits from a 32 bit field on 32bit glibc, for example.

Generally speaking, I'd recommend doing `%lld` for format and `(long long)value` instead.

This goes for all instances.

src/ui.c
```diff
@@ -170,7 +170,7 @@ file_finder_draw(struct file_finder *finder)
 
 	wmove(finder->win, finder->height - 1, 0);
 	wbkgdset(finder->win, get_line_attr(NULL, LINE_TITLE_FOCUS));
-	wprintw(finder->win, "[finder] file %d of %d", pos->lineno + 1, finder->lines);
+	wprintw(finder->win, "[finder] file %lu of %zu", pos->lineno + 1, finder->lines);
```
ericonr, line 173: In https://github.com/jonas/tig/pull/1085/files#diff-662ec1e8c57906cca33cf41b09c520a818a03e8c7454538e3529869b8c2b24b4R693 you're using `%u` for `line->lineno`. Do the `lineno` members have different types?

src/draw.c
```diff
@@ -322,7 +322,7 @@ draw_lineno_custom(struct view *view, struct view_column *column, unsigned int l
 		return false;
 
 	if (lineno == 1 || (lineno % interval) == 0) {
-		static char fmt[] = "%ld";
+		static char fmt[] = "%#u";
```
ericonr, line 325: I'm reasonably sure `u` can't use `#`. GCC even warns about it.

koutcher:
Thanks for the review @ericonr, I was sure we could use more than one pair of eyes on this kind of subject :-).

src/draw.c
koutcher: Actually # is just a placeholder, it is replaced with a digit one line below. I'll use %3u for clarity.

src/util.c
koutcher: Agreed, let's go the BSD way.

I actually took example on Git but I overlooked that they define their own timestamp_t type (though there was a clue as they use the unsigned version PRIuMAX).

src/ui.c
koutcher: Yes, I noticed that as too, lineno is unsigned int in line struct while it is unsigned long in position struct. For now, I only wanted to restore the concordance between formats and types, but I agree it could deserve attention as well.

PR #1088, opened 2021-02-23
Fix loop when wrapping line with ISO-8859-1 character

### LINKED ISSUE #1087: eats all memory and goes into swap
Recently I upgraded from tig-2.5.0 to tig-2.5.2. I can no longer view big commits: tig eats up all memory and goes into swap.

Steps to reproduce:
1) Clone e.g. [this repo](http://git.altlinux.org/people/at/packages/0ad.git).
2) In another terminal, run `top` and press `M` to sort by memory usage.
3) In yet another terminal, prepare `killall tig`.
4) Run `tig --all` and position to the second to last commit, the one with `0ad-0.0.24b-alpha-unix-build.tar.xz`.
5) Now press `Enter` to view the commit.
6) You'll only have a few seconds to peek at `top` and execute `killall tig`. After those few seconds, you won't be able to abort tig gracefully, it'll send the whole machine into swap.

The problem is exacerbated by the fact that tig traps signals. So by the time you realize that `C-C` doesn't work, and you want to open another terminal, it's already too late. I think it must be possible to trap signals in such a way that the first arrival resets the handler, so that the second `C-C` is a sure-fire way to kill the program.

I think I was able to bisect the problem to 484aa21a6f9ec6090a78b111eb429f9d922e25b9.

koutcher:
Fixes #1087

### DIFF
--- a/src/string.c
+++ b/src/string.c
@@ -214,72 +214,25 @@ unicode_width(unsigned long c, int tab_size)
 
 /* Number of bytes used for encoding a UTF-8 character indexed by first byte.
  * Illegal bytes are set one. */
-static const unsigned char utf8_bytes[256] = {
-	1,1,1,1,1,1,1,1, 1,1,1,1,1,1,1,1, 1,1,1,1,1,1,1,1, 1,1,1,1,1,1,1,1,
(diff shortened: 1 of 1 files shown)


krobelus:
Looks good; other places that handle utf8 are not affected

koutcher:
I managed to embed everything in utf8_length().

PR #1096, opened 2021-03-25
Avoiding null pointer dereference

binho1991:
This pull request aims to fix null pointer dereference.

[Null pointer dereference](https://cwe.mitre.org/data/definitions/476.html)

### DIFF
--- a/src/argv.c
+++ b/src/argv.c
@@ -221,11 +221,13 @@ argv_appendn(const char ***argv, const char *arg, size_t arglen)
 		return false;
 
 	alloc = strndup(arg, arglen);
+	if (!alloc)
+		die("Failed to allocate arg");
 
 	(*argv)[argc++] = alloc;
 	(*argv)[argc] = NULL;
 
-	return alloc != NULL;
+	return true;
 }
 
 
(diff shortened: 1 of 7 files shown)


koutcher:
@jonas, I started to review this PR from @binho1991 and have some suggestions, see https://github.com/koutcher/tig/commit/gh-1096-null-pointer-deref.

krobelus:
there are many other places where allocation can fail. Shouldn't we just use xmalloc/xcalloc like git does?

binho1991:
Hi @koutcher and @krobelus! Does anything still need to be done for the pull request to be accepted? Some points do not use xmalloc/xcalloc. Is there any difference in their use? If need be, I can change the pull request to use xmalloc and xcalloc. Thanks!

koutcher:
@binho1991, thank you, I don't think xmalloc/xcalloc would bring any benefit. You did a pretty good job at identifying the missing checks, only a few remained in argv_append. I completed my review in https://github.com/koutcher/tig/commit/4958efebb6c122745b63a4d5dab12610e28c4f2e (see src/argv.c and src/util.c), I believe it should be enough.

binho1991:
@koutcher thanks for the feedback.

PR #1098, opened 2021-04-07
Add TIG_EDITOR environment variable to configure editor

### LINKED ISSUE #889: Feature request: separate editor for git actions (blocking) vs files (non-blocking editor)
(This might be a little niche...)

It might make sense to allow different EDITOR selection for "edit file" action then for "edit commit message" action.
My reasoning here is that "edit commit" (and probably all of git actions that require editor) is a blocking action - git waits for the editing to complete before continuing. "Edit a file", OTOH, might be a long operation - open the file in my IDE in a separate window, and continue whatever it was I was doing in tig in this terminal - like maybe open a couple more files from the same diff.

(In my case, I use Notepad++ for "long edits" and Nano for "short" ones, because it's hard to setup N++ to "blocking" mode).

koutcher:
Closes #889

### DIFF
--- a/doc/tig.1.adoc
+++ b/doc/tig.1.adoc
@@ -235,6 +235,11 @@ TIG_NO_DISPLAY::
 	Open Tig without rendering anything to the terminal. This force Ncurses
 	to write to /dev/null. The main use is for automated testing of Tig.
 
+TIG_EDITOR::
+	The editor command to use when visiting files. This environment
+	variable overrides $GIT_EDITOR, $EDITOR and $VISUAL, so it allows
+	to use a different editor from the one Git uses.
+
 FILES
 -----
 '$XDG_CONFIG_HOME/tig/config'::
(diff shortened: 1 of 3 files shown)


doc/tig.1.adoc
```diff
@@ -235,6 +235,10 @@ TIG_NO_DISPLAY::
 	Open Tig without rendering anything to the terminal. This force Ncurses
 	to write to /dev/null. The main use is for automated testing of Tig.
 
+TIG_EDITOR::
+	The editor command. This environment variable overrides $GIT_EDITOR,
+	$EDITOR and $VISUAL.
```
krobelus, line 240: We could state the use case, even though it's fairly obvious. Maybe something like this:

The editor command to use when visiting files. This environment variable overrides $GIT_EDITOR,
$EDITOR and $VISUAL, so it allows to use a different editor from the one Git uses.

PR #1127, opened 2021-07-11
Allow :back to work across separate blame views

### LINKED ISSUE #1123: Allow `:back` to work across separate blame views
The blame view supports going back to the previous commit:
after pressing `b` on some commit in blame mode (to show the blame view starting from that commit), we can use `<` to go back.

I suggest that we extend this feature to work across blame views, where it would be more useful to me.

When I use `git blame` to find the origin of a line, I oftend need to traverse commits that just moved the line.
I'll describe what I usually with the of how to find the origin of the first line in README.adoc.

1. Run `tig blame 529182c -- README.adoc`
2. Press `<Enter>` to inspect the commit that added the first line (c91ba8f).
  Observe that it's just a cosmetic change, which we want to skip over.
3. Run `:272` to go to the place in the diff where first README-line was changed.
  Now we are on the deleted line `-Tig: text-mode interface for git`
4. Press `b` to go to the commit that introduced that line (b8ae934).
5. Proposed feature: now, pressing `<` (`back`) should take me back to the commit in the first blame-view. This doesn't happen because each view has a separate history.

I'm not sure how to implement this. I think there is already just one blame-view instance, so it should be enough to simply not delete the old blame-view history when creating a new one.

koutcher:
* stop clearing the history state when entering the blame view

* push a blame view history state before entering the diff view

* stop failing when trying to push an identical history state

Closes #1123

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -393,7 +393,7 @@ push_view_history_state(struct view_history *history, struct position *position,
 
 	if (state && data && history->state_alloc &&
 	    !memcmp(state->data, data, history->state_alloc))
-		return NULL;
+		return state;
 
 	state = calloc(1, sizeof(*state) + history->state_alloc);
 	if (!state)
--- a/src/blame.c
+++ b/src/blame.c
@@ -173,8 +173,6 @@ blame_open(struct view *view, enum open_flags flags)
 			free(blame->commit);
 	}
 
-	if (!(flags & OPEN_RELOAD))
-		reset_view_history(&blame_view_history);
 	string_copy_rev(state->history_state.id, view->env->ref);
 	state->history_state.filename = get_path(view->env->file);
 	if (!state->history_state.filename)
@@ -455,6 +453,8 @@ blame_request(struct view *view, enum request request, struct line *line)
 	enum open_flags flags = view_is_displayed(view) ? OPEN_SPLIT : OPEN_DEFAULT;
 	struct blame *blame = line->data;
 	struct view *diff = &diff_view;
+	struct blame_state *state = view->private;
+	struct blame_history_state *history_state = &state->history_state;
 
 	switch (request) {
 	case REQ_VIEW_BLAME:
@@ -475,6 +475,11 @@ blame_request(struct view *view, enum request request, struct line *line)
 		if (view_is_displayed(diff) &&
 		    !strcmp(blame->commit->id, diff->ref))
 			break;
+
+		if (!push_view_history_state(&blame_view_history, &view->pos, history_state)) {
+			report("Failed to save current view state");
+			break;
+		}
 
 		if (string_rev_is_null(blame->commit->id)) {
 			const char *diff_parent_argv[] = {


krobelus:
Thanks, this works great! I felt like it would be easy but I didn't understand enough of the view code.

> stop failing when trying to push an identical history state

Makes sense, this is required so we don't error on `tig blame 529182c -- README.adoc` Enter q Enter.

krobelus:
I found there is a similar use case for the diff view.

- `tig show 529182c -- README.adoc`
- `:23`
  - The cursor should be on the line after "The latest version is", that is, the deleted line containing `tig-2.5.3`
- Press `b` to enter the blame view.
  - What is a bit strange here is that the blame view shows the commit that
  added the line with `tig-2.5.4`. I expect it to show the commit that
  added the line with `tig-2.5.3`. This would be consistent with how we
  blame other deletions, so this might be an off-by-one error in `tig show`
- Press `d` to show the diff of commit tig-2.5.4, and `:23` to go to the same line (deleted `tig-2.5.3`).
- Press `b` to blame the same line again but this time we actually get to the
  `tig-2.5.3` commit.
- Press `d` to show the diff of commit tig-2.5.3. Now there is no nice way to get back to the diff of commit tig-2.5.4.
  - If I press `q` twice, I get back to the diff view which is still on the
  stack, but it shows commit tig-2.5.4. So I think we want to teach the diff
  view how to save/restore state, so `:back` or `q` could both go back to the earlier commit.

koutcher:
There is an issue when the diff view is the initial view. `argv_env.commit` is not set and the commit selection for blaming is done against `HEAD` instead of `529182c`. 

Here is a quick fix but I'm not completely sure it's the right way to do:
```patch
--- a/src/diff.c
+++ b/src/diff.c
@@ -34,6 +34,9 @@ diff_open(struct view *view, enum open_flags flags)
 	};
 	enum status_code code;
 
+	if (is_initial_view(view) && opt_rev_args)
+		string_copy_rev(argv_env.commit, opt_rev_args[0]);
+
 	diff_save_line(view, view->private, flags);
 
 	code = begin_update(view, NULL, diff_argv, flags);
```

krobelus:
I've openend #1135 with my attempt at a different fix. It's a bit hacky, I'll recheck it soon.

The other issue with the diff-view history would be solved if `:view-close` could restore the per-view history, like `:back` does for the blame view. The diff view doesn't have it's own history yet.
Currently, when in the blame view, a `:view-blame` will not push a view to the view stack, but only modify the blame view's history.
We could simplify this by always pushing a new blame-view on `:view-blame`. This would mean that every view on the view stack corresponds to (at most) one view state, which we can restore on `:view-close`. Of course this might make `:view-close` worse for the blame view.

src/view.c
```diff
@@ -393,7 +393,7 @@ push_view_history_state(struct view_history *history, struct position *position,
 
 	if (state && data && history->state_alloc &&
 	    !memcmp(state->data, data, history->state_alloc))
-		return NULL;
+		return state;
```
jonas, line 396: Good catch. Return the current stack when nothing changed?

jonas (approved):
I like the idea of keeping the history across "blame uses". The initial use case was mostly to be able to deep dive and drill down into "parent blame views". But more horizontal exploration also makes sense.

src/view.c
koutcher: Yes, otherwise reopening the same blame fails.

PR #1128, opened 2021-07-11
Log view fixes

koutcher:


src/log.c
```diff
@@ -68,8 +68,15 @@ log_open(struct view *view, enum open_flags flags)
...
 			"%(revargs)", "--no-color", "--", "%(fileargs)", NULL
 	};
+	enum status_code code;
 
-	return begin_update(view, NULL, log_argv, flags);
+	code = begin_update(view, NULL, log_argv, flags);
+	if (code != SUCCESS)
+		return code;
+
+	watch_register(&view->watch, WATCH_HEAD | WATCH_REFS);
```
krobelus, line 77: Changes look good, I just have a related question - I can't figure out how to make `set refresh-mode = after-command` work. It doesn't refresh after I run `:!echo` and type `q`.

koutcher: I must say that I don't use it either. Wondering if anyone does... WATCH_EVENT_AFTER_COMMAND seems to be trigged only after opening an editor or running a command requiring confirmation.

src/draw.c
```diff
@@ -543,12 +543,13 @@ view_column_draw(struct view *view, struct line *line, unsigned int lineno)
 		{
 			enum line_type type = line->type;
 			const char *text = column_data.text;
+			size_t indent;
```
jonas, line 546: Shouldn't this be initialized since we only set it to `get_graph_indent(text)` if `line->graph_indent` is non-zero?
```suggestion
			size_t indent = 0;
```

src/log.c
jonas: Hehm, yeah, the refresh code is really buggy unfortunately. It probably needs to be completely redesigned. Might be better to rip it out.

jonas (approved):

src/draw.c
koutcher: Both the initialisation and the use are conditioned by `line->graph_indent` but, agreed, it never hurts to initialise a variable in case someone wants to use it in the future and misses this condition.

PR #1135, opened 2021-07-24
Start blame on the commit passed to a "tig show" commandline

krobelus:
Fixes the first issue described in https://github.com/jonas/tig/pull/1127#issuecomment-884444332

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -635,6 +635,8 @@ diff_get_lineno(struct view *view, struct line *line, bool old)
 static enum request
 diff_trace_origin(struct view *view, struct line *line)
 {
+	struct line *commit_line = find_prev_line_by_type(view, line, LINE_COMMIT);
+	char id[SIZEOF_REV];
 	struct line *diff = find_prev_line_by_type(view, line, LINE_DIFF_HEADER);
 	struct line *chunk = find_prev_line_by_type(view, line, LINE_DIFF_CHUNK);
 	const char *chunk_data;
@@ -645,7 +647,7 @@ diff_trace_origin(struct view *view, struct line *line)
 	struct blame_header header;
 	struct blame_commit commit;
 
-	if (!diff || !chunk || chunk == line) {
+	if (!diff || !chunk || chunk == line || diff < commit_line) {
 		report("The line to trace must be inside a diff chunk");
 		return REQ_NONE;
 	}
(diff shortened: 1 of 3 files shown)


koutcher:
I found a better approach: get the commid id from the commit line in the diff. It also works when there are several commits in the diff view (e.g. `tig show --all`). See https://github.com/koutcher/tig/commit/6c77c61ddadc88f7887625ea147404fbc61a5ae5.

krobelus:
Yeah that seems better. I used %(commit) which works the same way except it also works for stash objects.
Unfortunately there's another regression with blame on stashes, see the stash test case.
Maybe we should just include the object-ID of a stashes in their diffs?

koutcher:
I don't think there is an option in `git stash show` to do that. We could use `git show` but I'm not sure we want to do that. Based on my initial proposal, we could fix the problem by using view->vid when there is no commit line (see https://github.com/koutcher/tig/commit/blame-initial-diff-view). I liked the way you used %(commit) but I don't see any easy way to fix the stash issue with it.

koutcher:
Based on your version, we could keep using view->vid and update it together with view->env->commit in the diff view:
```patch
diff --git a/src/diff.c b/src/diff.c
index 0ede47e..e6ef8cb 100644
--- a/src/diff.c
+++ b/src/diff.c
@@ -687,9 +687,9 @@ diff_trace_origin(struct view *view, struct line *line)
        }

        if (chunk_marker == '-')
-               string_format(ref, "%s^", view->env->commit);
+               string_format(ref, "%s^", view->vid);
        else
-               string_copy(ref, view->env->commit);
+               string_copy(ref, view->vid);

        if (!diff_blame_line(ref, file, lineno, &header, &commit)) {
                report("Failed to read blame data");
diff --git a/src/pager.c b/src/pager.c
index b570de2..c154c2b 100644
--- a/src/pager.c
+++ b/src/pager.c
@@ -184,6 +184,7 @@ pager_select(struct view *view, struct line *line)

        if (line->type == LINE_COMMIT) {
                string_copy_rev_from_commit_line(view->env->commit, text);
+               string_copy_rev_from_commit_line(view->vid, text);
                if (!view_has_flags(view, VIEW_NO_REF))
                        string_copy_rev(view->ref, view->env->commit);
        }
```

krobelus:
Ok falling back to view-vid if there is no commit line seems to work.
I think this is a workaround because `view->env->commit` should also be set if we go backwards.

koutcher:
Yes and no. When we fix view->env->commit (let's put it on the todo list, we can reuse bits from the log view), we will still need to look for a commit line to know when we have to use view->vid as view->env->commit can be overwritten by other views.

I noticed test/blame/initial-diff-test fails on my Mac because of `sed -i`. `sed -i.old -n /asInstanceOf/p *.screen` fixes it.

krobelus:
> as view->env->commit can be overwritten by other views.

What is an example of that overwriting? Shouldn't view->env->commit (aka %(commit)) always be the commit of the thing at the cursor?
Since :view-blame blames at the cursor, view->env->commit should then be the correct one.
Otherwise my commit message is slightly off.

koutcher:
This was typically what happened in your scenario with a stash. When the diff view is opened on the stash, view->env->commit is initialized from the stash view, but it is later overwritten in the blame view and when coming back to the diff view, there is nothing to set view->env->commit back as there is no commit line.

krobelus:
Ah right, so we do just need to fix view->env->commit but that probably requires a commit line.

Issue #628, opened 2017-06-29
[wishlist] "browsing" history in the "main view"


yarikoptic:
Disclaimer: new to tig (saw/tried it but then forgot -- now starting to use again since it is really nice. Thanks!)

I got to love PyCharm feature to be able navigate me to the previous locations in the code which I have recently visited and/or modified.

tig allows you to quickly jump to a parent (`,`), but I do not see any easy way to get back to the commit where you came from.  Altogether I felt that it might be cool to have an easy way to at least "get back" (could be the unused `<` in the main view)

PR #1138, opened 2021-08-12
Enable jumping back in main view

kempniu:
Hi there,

First, thank you for all the work on this awesome tool.

The one feature that I have been missing is the ability to go back to the previously selected commit after jumping to its parent.  I find that to be really useful for git repositories which are "heavily non-linear" (like the Linux kernel tree).

I see this idea has already been suggested in #628 and [also](https://github.com/jonas/tig/pull/506#issuecomment-240302791) in #506 (which implemented #388), so I took a stab at implementing it.

This PR looks almost too simple to be right, but it seems to work well for me :-)  I would be happy to work on it further if the approach I chose is not the preferred one.

Hope this helps!

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -32,6 +32,8 @@
 
 DEFINE_ALLOCATOR(realloc_reflogs, char *, 32)
 
+static struct view_history main_view_history = { 0 };
+
 bool
 main_status_exists(struct view *view, enum line_type type)
 {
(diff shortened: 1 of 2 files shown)


krobelus:
Wow I never thought of using `:parent` in main view - LGTM

koutcher:
Thanks, this makes the navigation in a messy git history really much easier.

koutcher (approved):

typebrook:
Works like a charm! Thank you @kempniu

PR #1139, opened 2021-08-14
Autoscroll colorify compiler output

psprint:
Hi,
I think that this PR implements quite a valuable features, which all converge to a very useful use case (explained at the end). The features are:

## The patches
 
1. **Autoscroll** of the pager view – quite a missed feature, with a protection against overactivity – the view has to be placed at the bottom for the autoscroll to occur – i.e.: it's safe to scroll up and have a stable view of what's going on.

2. **Colorizing** of the compiler's errors and warnings (and notes) in pager view. <br/>The errors are to be browsed in the use case explained below, so they are being colorized first here by this patch.

## The use case

The two patches open a way for a configuring of TIG as a **comfortable IDE-like errors round-trip fixing tool**, by:

1. Configuring the `'n'` key in pager to search for lines with "error:|warning:|note:" regex, via.:
    ```
    bind pager n :/warning:|error:|note:
    ```
2. Configuring Return in pager to open the editor on the line number extracted from the current error/warning/note, via:
    ```
    bind pager <Enter> !sh -c "~/github/tig/contrib/make-line-open.zsh '%(text)'"
    ```

3. Optionally, configuring `'E'` to run `:!make`, via:
    ```
    bind generic E :!make
    ```

4. Setting the `compiler-msg` color (the only thing that's problematic - I don't know how to make a default value for it), via:
    ```
    color compiler-msg yellow default bold
    ```

Then what's possible is the following use case: [asciicast](https://asciinema.org/a/430460).

## Helper script body

The `make-line-open` script is as follows:

```zsh
#!/usr/bin/env zsh

emulate zsh -o extended_glob

if [[ $1 == (#b)([^:]##):([0-9]##):* ]]; then
    [[ -z $EDITOR ]] && EDITOR=mcedit

    $EDITOR +$match[2] **/$match[1]
fi
```

## Summary

The patches are quite simple, only 9 and 21 lines and follow the coding style. I think that the patches are useful also without/outside the use case – having pager view autoscrolled, for example, is an useful thing.

### DIFF
--- a/include/tig/line.h
+++ b/include/tig/line.h
@@ -22,6 +22,7 @@ struct ref;
  */
 
 #define LINE_INFO(_) \
+        _(COMPILER_MSG,     ""), \
 	_(DIFF_HEADER,		"diff --"), \
 	_(DIFF_DEL_FILE,	"--- "), \
 	_(DIFF_ADD_FILE,	"+++ "), \
--- a/src/diff.c
+++ b/src/diff.c
@@ -302,12 +302,33 @@ diff_common_highlight(struct view *view, const char *text, enum line_type type)
 bool
 diff_common_read(struct view *view, const char *data, struct diff_state *state)
 {
+        const char *regex_txt;
+        regex_t regex;
+        regmatch_t pmatch[10];
+	int regex_flags = REG_EXTENDED, regex_err;
 	enum line_type type = get_line_type(data);
 
 	/* ADD2 and DEL2 are only valid in combined diff hunks */
 	if (!state->combined_diff && (type == LINE_DIFF_ADD2 || type == LINE_DIFF_DEL2))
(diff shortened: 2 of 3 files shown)


krobelus (commented):
Thanks for sharing but this seems quite specific to your use case.
The autoscroll feature should not be active by default - Tig is emulating `less`.
Autoscrolling makes a `:!git log` deviate from `less`.

koutcher:
Thanks for your contribution. I tend to agree with Johannes on the specificity of the use case, especially we shall not embed hardcoded patterns. I'll leave it to Jonas to decide.

psprint:
In a few days I'll provide an updated patch that will address the points, e.g.: will use an option for the regex.

koutcher:
Superseded by #1223 and #1249.

Issue #542, opened 2016-11-20
Supporting diff-so-fancy in diff view


Liooo:
It'd be nice if tig supported [diff-so-fancy](https://github.com/so-fancy/diff-so-fancy), as previously mentioned in #313.

diego898:
just want to bump this!

EnCey:
Is there a workaround for this? 
`diff-highlight` is supported apparently, but I didn't quite get why diff-so-fancy isn't or what needs to be done to make it work

yilazius:
Any updates on this issue? Did somebody get it working?

alok:
I'm curious about this too. I really like the clean view d-s-f gives.

mathomp4:
I just learned about `diff-so-fancy` myself. Bump and :+1:

graves501:
diff-so-fancy support would be awesome! bump :)

patrickelectric:
This would be awesome !

gwellner:
Oh i need this Feature !

andebjor:
I have looked into adding this support today. Unfortunately, I have not succeeded, and I currently think that I will take quite some effort to implement this support properly.

The `diff-so-fancy` tool works as a filter of the `git diff` output (as does `diff-highlight` which `tig` has support for). The filter does basically these things:

1. Tweaks the header of each touched file,
2. Tweaks the hunk indicators to be more human readable,
3. Strips the leading `+` and `-` signs,
4. Marks empty lines, and
5. Adds a fair amount of coloring to different elements of the diff.

It is the added coloring that makes this a hard task. As @jonas mentions in #313,`tig` was only taught very basic parsing of ANSI control codes when support for `diff-highlight` was added. Actually, `tig` has a quite advanced support for coloring of the diffs built-in, and uses the `--no-color` flag when retrieving the diff, and proceeds to add colors afterwards.

The same could perhaps be done with `diff-so-fancy`, except that I have not found any way to suppress adding of the ANSI control characters to its output for many elements of the diff. (That could be a pull request to that tool, however, if desired.)

At my current level of understanding of the situation I see these ways forward for this feature:

1. Teach `tig` to parse ANSI control characters, or
2. Implement native support for the other items in the list of features that `diff-so-fancy` supports (items 1--4 in the list above basically).

[...]

PR #1140, opened 2021-08-17
support delta along with ANSI support

ulwlu:
https://github.com/jonas/tig/issues/542#issuecomment-899094741

- All test passes.
- I haven't seen segmentation faults with newest versions of MacOS Big Sur, iTerm2, and delta. ncurses version is 6.2.
- It looks like no performance issues.
- I've only tested it with Terminal.app and iTerm on MacOS.

ss 0003-08-17 at 19 11 58
ss 0003-08-17 at 19 12 36

### how to use delta in tig

It requires ncurse 6.1 or higher.

```bash
# if you don't have ncurses
curl -O ftp://ftp.gnu.org/gnu/ncurses/ncurses-6.2.tar.gz
tar -xzvf ncurses-6.2.tar.gz
cd ncurses-6.2
./configure --prefix=${where you like} --enable-pc-files --with-pkg-config-libdir=/usr/local/bin/pkgconfig --enable-sigwinch --enable-symlinks --enable-widec --with-shared --with-gpm=no --without-ada --enable-ext-colors # i set the path as /usr/local which is default
sudo make
sudo make install

cd ${path to tig of the branch of above PR}
make configure
./configure LDFLAGS=-L${where you download ncurses}/lib CPPFLAGS=-I{where you installed ncurses}/include  # i set the path as /usr/local/opt/ncurses which is installed by ncurses
make

echo "set diff-highlight = \"delta\"" >> ~/.tigrc

./src/tig
```

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "editor.tabSize": 8,
+  "editor.insertSpaces": false,
+  "files.associations": {
+    "*.h": "c"
+  }
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,0 @@
-{
-  "editor.tabSize": 8,
-  "editor.insertSpaces": false,
-  "files.associations": {
-    "*.h": "c"
-  }
-}
--- a/src/ansi.c
+++ b/src/ansi.c
@@ -82,7 +82,6 @@ draw_ansi(struct view *view, int *ansi_num, char **ansi_ptrs, int max_width, siz
 
 		if (view->curline->selected) {
 			draw_ansi_line(view, ansi_end_ptr, after_ansi_len, &skip, &cur_width, &widths_of_display);
-			cur_width += widths_of_display;
 			continue;
 		}
 
@@ -173,7 +172,6 @@ draw_ansi(struct view *view, int *ansi_num, char **ansi_ptrs, int max_width, siz
 		wattrset_by_ansi_status(view, &cur_ansi_status);
 
 		draw_ansi_line(view, ansi_end_ptr, after_ansi_len, &skip, &cur_width, &widths_of_display);
-		cur_width += widths_of_display;
 
 		free(ansi_code);
 		ansi_code = NULL;
@@ -204,6 +202,8 @@ draw_ansi_line(struct view *view, char *ansi_end_ptr, int after_ansi_len, size_t
 	} else {
 		waddnstr(view->win, ansi_end_ptr, after_ansi_len);
 	}
+
+	*cur_width += *widths_of_display;
 }
 
 void
--- a/src/ansi.c
+++ b/src/ansi.c
@@ -85,92 +85,81 @@ draw_ansi(struct view *view, int *ansi_num, char **ansi_ptrs, int max_width, siz
 			continue;
 		}
 
+		// ncurses can't handle multiple attribute such as BOLD & UNDERLINE.
+		// If input-ansi has "\033[1;4" we'll give priority to the latter one.
+		char *saveptr;
 		char *ansi_code = malloc(sizeof(char) * (ansi_code_len + 1));
 		strncpy(ansi_code, text + 2, ansi_code_len);
 		ansi_code[ansi_code_len] = '\0';
-		if (strchr(ansi_code, ';') == NULL) {
-			if (strcmp(ansi_code, "0") == 0)
+		char *ansi_code_part = strtok_r(ansi_code, ";", &saveptr);
+		while (ansi_code_part != NULL) {
+			if (strcmp(ansi_code_part, "0") == 0) {
+				cur_ansi_status.fg = 256;
+				cur_ansi_status.bg = 256;
 				cur_ansi_status.attr = A_NORMAL;
-			if (strcmp(ansi_code, "1") == 0)
-				cur_ansi_status.attr = A_BOLD;
-			if (strcmp(ansi_code, "2") == 0)
-				cur_ansi_status.attr = A_DIM;
-			if (strcmp(ansi_code, "3") == 0)
-				cur_ansi_status.attr = A_ITALIC;
-			if (strcmp(ansi_code, "4") == 0)
-				cur_ansi_status.attr = A_UNDERLINE;
-			if (strcmp(ansi_code, "5") == 0)
-				cur_ansi_status.attr = A_BLINK;
-			if (strcmp(ansi_code, "6") == 0)
-				cur_ansi_status.attr = A_BLINK; // This is supposed to be faster than normal blink, but ncurses doesn't have any way to achieve.
-			if (strcmp(ansi_code, "7") == 0)
-				cur_ansi_status.attr = A_REVERSE;
-			if (strcmp(ansi_code, "8") == 0)
-				cur_ansi_status.attr = A_INVIS;
-			if (strcmp(ansi_code, "9") == 0)
-				// This is supposed to be strikethrough, but ncurses doesn't have any way to achieve.
-			if (strcmp(ansi_code, "30") == 0)
-				cur_ansi_status.fg = COLOR_BLACK;
-			if (strcmp(ansi_code, "31") == 0)
-				cur_ansi_status.fg = COLOR_RED;
-			if (strcmp(ansi_code, "32") == 0)
-				cur_ansi_status.fg = COLOR_GREEN;
-			if (strcmp(ansi_code, "33") == 0)
-				cur_ansi_status.fg = COLOR_YELLOW;
-			if (strcmp(ansi_code, "34") == 0)
-				cur_ansi_status.fg = COLOR_BLUE;
-			if (strcmp(ansi_code, "35") == 0)
-				cur_ansi_status.fg = COLOR_MAGENTA;
-			if (strcmp(ansi_code, "36") == 0)
-				cur_ansi_status.fg = COLOR_CYAN;
-			if (strcmp(ansi_code, "37") == 0)
-				cur_ansi_status.fg = COLOR_WHITE;
-			if (strcmp(ansi_code, "40") == 0)
-				cur_ansi_status.bg = COLOR_BLACK;
-			if (strcmp(ansi_code, "41") == 0)
-				cur_ansi_status.bg = COLOR_RED;
-			if (strcmp(ansi_code, "42") == 0)
-				cur_ansi_status.bg = COLOR_GREEN;
-			if (strcmp(ansi_code, "43") == 0)
-				cur_ansi_status.bg = COLOR_YELLOW;
-			if (strcmp(ansi_code, "44") == 0)
-				cur_ansi_status.bg = COLOR_BLUE;
-			if (strcmp(ansi_code, "45") == 0)
-				cur_ansi_status.bg = COLOR_MAGENTA;
-			if (strcmp(ansi_code, "46") == 0)
-				cur_ansi_status.bg = COLOR_CYAN;
-			if (strcmp(ansi_code, "47") == 0)
-				cur_ansi_status.bg = COLOR_WHITE;
-		} else {
-			char *token = malloc(sizeof(char) * (ansi_code_len + 1));
-			strcpy(token, ansi_code);
-			char *ansi_code_part = strtok(token, ";");
-
-			while (ansi_code_part != NULL) {
-				char *color_method_mark = strtok(NULL, ";");
-				if (strcmp(color_method_mark, "5") == 0) {
-					char *c256 = strtok(NULL, ";");
-					if (strcmp(ansi_code_part, "38") == 0)
-						cur_ansi_status.fg = atoi(c256);
-					if (strcmp(ansi_code_part, "48") == 0)
-						cur_ansi_status.bg = atoi(c256);
-				}
-				// WONTFIX: You can't init_color with numerous RGB code in ncurses.
-				// I decided to force delta users to use "true-color = never" when using tig,
-				// so the process never comes to this condition.
-				// I leave the code for someone who wants to implements in the future.
-				// if (strcmp(color_method_mark, "2") == 0) {
-					// char *r = strtok(NULL, ";");
-					// char *g = strtok(NULL, ";");
-					// char *b = strtok(NULL, ";");
-				// }
-				ansi_code_part = strtok(NULL, ";");
 			}
-			free(token);
-			token = NULL;
+			if (strcmp(ansi_code_part, "1") == 0)
+				cur_ansi_status.attr = A_BOLD;
+			if (strcmp(ansi_code_part, "2") == 0)
+				cur_ansi_status.attr = A_DIM;
+			if (strcmp(ansi_code_part, "3") == 0)
+				cur_ansi_status.attr = A_ITALIC;
+			if (strcmp(ansi_code_part, "4") == 0)
+				cur_ansi_status.attr = A_UNDERLINE;
+			if (strcmp(ansi_code_part, "5") == 0)
+				cur_ansi_status.attr = A_BLINK;
+			if (strcmp(ansi_code_part, "6") == 0)
+				cur_ansi_status.attr = A_BLINK; // This is supposed to be faster than normal blink, but ncurses doesn't have any way to achieve.
+			if (strcmp(ansi_code_part, "7") == 0)
+				cur_ansi_status.attr = A_REVERSE;
+			if (strcmp(ansi_code_part, "8") == 0)
+				cur_ansi_status.attr = A_INVIS;
+			if (strcmp(ansi_code_part, "9") == 0)
+				// This is supposed to be strikethrough, but ncurses doesn't have any way to achieve.
+			if (strcmp(ansi_code_part, "30") == 0)
+				cur_ansi_status.fg = COLOR_BLACK;
+			if (strcmp(ansi_code_part, "31") == 0)
+				cur_ansi_status.fg = COLOR_RED;
+			if (strcmp(ansi_code_part, "32") == 0)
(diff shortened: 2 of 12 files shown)


ulwlu:
This now works.

ss 0003-08-17 at 19 11 58

ss 0003-08-17 at 19 12 36

hasufell:
Great stuff.

How is the highlighting expected to work? I found it a bit surprising:

Peek 2021-08-17 17-36

hasufell:
I also seem to get some line skew with very long lines:

Peek 2021-08-17 17-51

ulwlu:
Hi hasufell, thank you for catching. I'll fix those.

ronjouch:
@ulwlu hi, here to respond to your call for tests at https://github.com/jonas/tig/issues/542#issuecomment-900179550 .

Your branch compiles and works for me (Arch Linux / [Kitty](https://sw.kovidgoyal.net/kitty/) 0.23.1, ncurses 6.2, delta 0.8.3) 🙂.

In addition to the weird line highlight already reported, I was surprised to see delta-within-tig uses a gray background. I'm not sure why, as pure delta outside of tig uses a black background. My `delta` config is empty, and my `~/.tigrc` is empty except for the single line `set diff-highlight = "delta"`.

Screencast below (pure delta at the beginning from 0:00 to 0:07, delta-in-tig afterwards):

https://user-images.githubusercontent.com/522085/129816178-73f18a78-b889-4aae-ab09-73d91c7a605a.mp4

ulwlu:
@ronjouch Hi, thank you for catching and details. I'm still not sure if this is caused by coloring problem, but I'm thinking delta gave some background color info with '--trur-color ＝never' that is not close to real true-color.

I'll check it anyway. Thank you very much

ronjouch:
> Thank you very much

Much obliged; thank you for the work on this!

> I'm still not sure if this is caused by coloring problem, but I'm thinking delta gave some background color info with '--trur-color ＝never' that is not close to real true-color.

@ulwlu I did a bit more monkeying around with kitty color configuration. The problem doesn't appear by default. But with my Kitty color config (pasted below), it does.

```
#: The color table {{{

#: The 256 terminal colors. There are 8 basic colors, each color has a
#: dull and bright version, for the first 16 colors. You can set the
#: remaining 240 colors as color16 to color255.

color0 #555753
color8 #777777
#: black

color1 #ef2929
color9 #cc0000
#: red

color2  #8ae234
color10 #4e9a06
#: green

color3  #fce94f
color11 #c4a000
#: yellow

color4  #729fcf
color12 #3465a4
#: blue

color5  #ad7fa8
color13 #75507b
#: magenta

color6  #34e2e2
color14 #06989a
#: cyan

color7  #eeeeec
color15 #d3d7cf
#: white
```

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "editor.tabSize": 8,
+  "editor.insertSpaces": false,
+  "files.associations": {
+    "*.h": "c"
+  }
+}
```
pablospe, line 7: I believe this file shouldn't be part of the PR

pablospe:
Another details. Have you tried **side-by-side** mode? It seems to work but only until 80 chars. So, I was thinking to use `delta --width 250`, but this doesn't seem to work: `set diff-highlight = "delta --width 250"`. How do you pass parameters?

I've also observed a weird highlighting, gray background. I am using konsole.

pablospe:
Also noticed that when there are many files and you press enter to the filename (in the commit header), it doesn't jump to beginning of the file in the diff. Instead I received this error: 
```
[diff] Press '<Enter>' to jump to file diff - line 11 of 88                                                                                                                                                                                 
Failed to find file diff
```

ulwlu:
@hasufell @ronjouch 

Hi. I've successfully fixed those, but I've come up with a better design. So I've decided to refactor it, which should take about a week.

@pablospe 

Tig can't accept formatter options. I can fix that but it will be on another PR since it has some impact on another function. Currently could you set config locally?

> I believe this file shouldn't be part of the PR

I knew what you want to say, but it would be a pain to develop if I separate PR for the vscode settings, so I'm doing this. If jonas tells about it, I'll separate it, but for now I won't.

> it doesn't jump to beginning of the file in the diff

Thank you for catching. I'll fix this.

pablospe:
> Tig can't accept formatter options. I can fix that but it will be on another PR since it has some impact on another function. Currently could you set config locally?
> 
> >

Apologies, I didn't understand. What do you mean by set config locally? Having an alias or something like that to avoid options?

ulwlu:
@pablospe 

Sorry for the lack of explanation, please do the following at this time.

```bash
git config delta.width 250
```

From your explanation, I think you don't want to set it to global. However, tig does not accept options now, so please set locally like this inside repositories that you use tig. Currently tig searches only executable binary, so alias doesn't work either. I will probably fix it to accept options if jonas agreed.

Also, in the original issue I asked what to do with the side-by-side option inside tig. If you have time, please feel free to post something like "allow tigrc to accept options and allow users to set width freely", so that someone may post an another idea.

https://github.com/jonas/tig/issues/542#issuecomment-899094741

pablospe:
Also another problem I found is that in navigation mode for delta doesn't work:
`n and N keybindings to move between files in large diffs, and between diffs in log -p views (--navigate)`. See in delta documentation ([here](https://github.com/dandavison/delta#deltas-main-features-are))

I am just reporting the problems I found. Thanks a lot for the fantastic work, @ulwlu!

krobelus:
I haven't checked the implementation details but this looks a bit weird on light terminals like gnome-terminal or xterm because it draws dark backgrounds.

ulwlu:
@krobelus Hi. Yes that occured because of here.

https://github.com/jonas/tig/pull/1140/files#diff-3cd1efc51226a700b579928f3ee8477ed1a05715a60b2d6e0b1fc54e63969581R50


I noticed that and fixed it into default_bg. Thank you for catching.

P.S.
Currently I'm fixing a bit of functions which pablospe mentioned.

pablospe:
> Currently I'm fixing a bit of functions which pablospe mentioned and more else.

Maybe some of the changes can be separated in other PRs and add basic `delta` support on master? Even if it is not perfect, it would be good to have the delta support.

ulwlu:
@pablospe sorry to be late. I was busy with my work and time flies so fast. That sounds great, I wanna add at least one more feature then.

- [x] handle delta output and draw correct color
- [x] draw correct current cursor
- [x] calculate the correct line length ( so that the line won't be overwrapped)
- [ ] handle correct pager <- i want to add at least this one

These can be called basic feature, and we can add some features such as jumping to the file hunks.

I probably can add in this year

ronjouch:
> @ronjouch I don’t have kitty env but I think I have fixed your case.

@ulwlu thanks, fix confirmed! I'll be running 629a12e and will report new problems here.

pablospe (approved):

pablospe:
Looks good to me! Thanks for the great job, @ulwlu! 
@jonas Who can approve this PR to be merged?

ulwlu:
thank you pablospe for your review!

probably @koutcher would you review this PR?

mbsulliv:
I've compiled 5ae769b and am having a segmentation fault on Linux.  I compiled with "make all-debug" and then ran tig via gdb and the crash is happening in draw_ansi() on line 151: the `color_method_mark` variable is null.

The crash seems related to lines like the following in my ~/.gitconfig:
`minus-emph-style = bold red 52`

If a line like that is present (i.e. a line with a background color), then the crash happens pretty reliably.  Deleting the background color (i.e. the 52) from that line makes the crash go away.  Note that even if the crash goes away, I still see some erratic behavior where delta sometimes doesn't display the diff correctly.

ulwlu:
@mbsulliv 
thank you for catching, i confirmed the bug.
It happens with the following string example, 

```
^[[34m^[[38;5;88m 28 ^[[34m⋮^[[38;5;28m    ^[[34m│^[[31mvoid draw_ansi(struct view *view, int *ansi_num, char **ansi_ptrs);^[[0m
```
.... because i should add the null check before the exact line where you pointed out `color_method_mark`.

this should be covered, i appreciate your catching. I'll fix and add a test as well. (until next Wednesday)

src/ansi.c
```diff
@@ -218,4 +207,27 @@ wattrset_by_ansi_status(struct view *view, struct ansi_status* cur_ansi_status)
 	wattr_set(view->win, cur_ansi_status->attr, id, NULL);
 }
 
+short
+convert_ansi_into_256_color(char *save_ptr) {
```
mbsulliv, line 211: I believe `save_ptr` needs to be a `char**` here because `strtok_r` modifies the value of the pointer.  As it is currently written, when the value of `save_ptr` gets modified, the changed value isn't passed back into the calling function.

I discovered this problem when it tripped over the following ANSI code: `\033[38;5;8m`.  The "38" triggered us to enter this function, the "5" was handled here, but upon returning, the "5" was parsed again and it caused all my code to start blinking.  It was pretty funny actually ;)

I tested this change on my side and things started behaving correctly.

src/ansi.c
```diff
@@ -85,92 +85,81 @@ draw_ansi(struct view *view, int *ansi_num, char **ansi_ptrs, int max_width, siz
...
-			if (strcmp(ansi_code, "44") == 0)
+			if (strcmp(ansi_code_part, "44") == 0)
 				cur_ansi_status.bg = COLOR_BLUE;
-			if (strcmp(ansi_code, "45") == 0)
+			if (strcmp(ansi_code_part, "45") == 0)
 				cur_ansi_status.bg = COLOR_MAGENTA;
-			if (strcmp(ansi_code, "46") == 0)
+			if (strcmp(ansi_code_part, "46") == 0)
 				cur_ansi_status.bg = COLOR_CYAN;
-			if (strcmp(ansi_code, "47") == 0)
```
mbsulliv: Is it intentional to delete the "47" code?

mbsulliv (changes requested):

src/ansi.c
ulwlu: You are correct, sorry for my mistake. Fixed here [5093bb1](https://github.com/jonas/tig/pull/1140/commits/5093bb1f89ec4309956fd4101e46aa64cce485a8).

This could be an excuse, but the reason why I didn't notice is that iTerm and VSCode-Terminal doesn't support BLINK.... And Macbook-builtin Terminal does support. Now I confirmed the bug and fixed. 
https://unix.stackexchange.com/questions/151673/attrona-blink-using-ncurses-does-nothing 

Thank you very much for catching again.

src/ansi.c
ulwlu: Sorry, it's just my mistake. 
Thanks for catching, Fixed here b86f198 .

fengyichui:
Hi,
In diff view, the file jump feature does not work properly and show error message: "Failed to find file diff"

ulwlu:
@fengyichui 
please see the comment above, this supports the basic feature. 

```
Maybe some of the changes can be separated in other PRs and add basic delta support on master? Even if it is not perfect, it would be good to have the delta support.
```

fengyichui:
@ulwlu  Sorry, I don't quite understand, do you mean it's already supported or need other PRs ?

ulwlu:
it needs other PR.

fengyichui:
@ulwlu  Thanks for your answer.

edubxb:
Looking forward to seeing this merged, awesome, thanks!

crokobit:
Can we merge this?

FdelMazo:
Hey @koutcher, any chance we can get this merged? I'd love this feature

ulwlu:
Thank you @koutcher for reviewing.
Your personal time should always come first, so never mind.

I think your considerations are all reasonable and I want to get started on them right away (especially the 256x256 fixed array, which I thought same thing when I created it. I didn't see any performance degradation in memory or speed after creation, but I'll look again to see if there are any other workarounds).

Actually, I will start working for a new company next week and will be a bit busy, so I will start next weekend,
but in terms of below question

> all these if (strcmp(... without elses, is it on purpose?

Yes. I didn't put else where else is not necessary.

Thanks again for taking the time to review this.

koutcher:
@ulwlu, please do. As I'm not a delta or any other fancy diff user, I'll create a feature branch to allow contributions until it is mature enough to be merged. Thanks again for your time.

ulwlu:
@koutcher 
just a note to let you know, I've already sent you a patch in email 12 days ago. (from ooulwluoo@gmail.com)
you don't need to reply this if you've noticed it.

ijoseph:
For Apple Silicon (M1)/ Homebrew, I had to

```
./configure  LDFLAGS="-L/opt/homebrew/opt/ncurses/lib" CPPFLAGS="-I/opt/homebrew/opt/ncurses/include"  PKG_CONFIG_PATH="/opt/homebrew/opt/ncurses/lib/pkgconfig"
```


based on 

```
╰─ brew info ncurses
==> ncurses: stable 6.3 (bottled) [keg-only]
Text-based UI library
https://invisible-island.net/ncurses/announce.html
/opt/homebrew/Cellar/ncurses/6.3 (3,968 files, 9.6MB)
  Poured from bottle on 2022-03-17 at 16:24:05
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/ncurses.rb
License: MIT
==> Dependencies
Build: pkg-config ✔
==> Caveats
ncurses is keg-only, which means it was not symlinked into /opt/homebrew,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

If you need to have ncurses first in your PATH, run:
  echo 'export PATH="/opt/homebrew/opt/ncurses/bin:$PATH"' >> ~/.zshrc

For compilers to find ncurses you may need to set:
  export LDFLAGS="-L/opt/homebrew/opt/ncurses/lib"
  export CPPFLAGS="-I/opt/homebrew/opt/ncurses/include"

For pkg-config to find ncurses you may need to set:
  export PKG_CONFIG_PATH="/opt/homebrew/opt/ncurses/lib/pkgconfig"
```

FdelMazo:
I'm not entirely sure. Does this PR helps towards adding `--color-moved` support? Issue #890

vitornesello:
This MR is pretty cool. Does anyone know when it will be released?

aberres:
Someone else just being here to ask if there are plans to get this in  😇

pablospe:
In the meanwhile, I ended up using this shortcuts (in the .tigrc):

```
# Open delta with git show.
bind generic d >sh -c "DELTA_PAGER='less -RKc' git show %(commit)"

# https://github.com/jonas/tig/issues/542#issuecomment-1242039892
bind generic D @sh -c "\
  ( \
    tmux has-session -t \".{last}\" \
    && tmux respawn-pane -t \".{last}\" -k 'LESS= DELTA_PAGER=\"less -R\" git show %(commit)' \
  ) \
  || tmux split-window -l 50% 'LESS= DELTA_PAGER=\"less -R\" git show %(commit)'"
```

First shortcut `d` is a generic to see the diff in full-screen. If you are inside tmux, you could also try the `D` shortcut, it is more convenient since it opens a new pane and you can see the list of commits too, but usually I end up using `d` in general because it is full-screen and I see more. What it is missing is a way to auto-detect if you are inside tmux, and select the proper behavior, rather than having two shortcuts. 

Obviously this is PR is a better solution, but for now this is a workaround that works for me. I hope this PR get merge soon.

pablospe:
Can this PR be re-open? @koutcher @jonas 
(the author account was deleted)

Alternative I could open a new PR from my github fork. I will wait some time before creating the PR myself, in case this PR could be re-open easily .

koutcher:
@pablospe, GH does not allow to reopen a PR from a deleted account. The latest status from the original author is already accessible in the ansi-support branch: https://github.com/jonas/tig/tree/ansi-support.

pablospe:
In that case, I will create a new PR pointing: https://github.com/jonas/tig/tree/ansi-support.

Few questions:
    * From where is the [Rework after review](https://github.com/jonas/tig/commit/d2b29cd2d424ccdd78ef080bd6b8e184847ce909) commit is coming from, it doesn't seem to be part of this PR.
    * What would be the next steps to get this PR in?

adamency:
Hi @pablospe, I've seen your fork https://github.com/pablospe/tig.

Can it be used as is (i.e. simply replacing the `tig` executable) to be able to use `delta` in tig ?

adamency:
I'm leaving this comment here for people who want to try delta in `tig` now. (The author) @uwlwu's instruction were not very precise.

## Installation Instructions

```
git clone https://github.com/jonas/tig/ 
cd tig
# Checkout branch of the project which contains the PR additions
git checkout ansi-support
make configure
# `ncurses` is almost always installable by your package manager, and almost always already installed on your system as it is needed for many TUI programs. The paths given here are the ones used by the package manager.
./configure LDFLAGS="-L/usr/lib" CPPFLAGS="-I/usr/include"
# This will install `tig` to ~/.local/bin, the recommended user path by the XDG specs. If needed, replace it by the path you want but leave out the `/bin/` part as this will be added by `make` already.
make prefix=$HOME/.local/
make install prefix=$HOME/.local/
# Configure `tig` to use `delta` in diffs
echo "set diff-highlight = \"delta\"" >> ~/.tigrc
```

This should work. However I have to note, some commits are displaying with the normal output instead of `delta`'s one. I don't know what's going wrong in theses cases.

pablospe:
Let's move the discussion to:
**PR https://github.com/jonas/tig/pull/1298:** [Support delta along with ANSI support (continuation)](https://github.com/jonas/tig/pull/1298) 

**Notes**: 
1. Why? GH does not allow to reopen a PR from a deleted account.
2. @koutcher I didn't have permission to create a new PR from https://github.com/jonas/tig/tree/ansi-support, so I did it from my fork, hope this is ok with you. What are the next steps?

PR #1143, opened 2021-09-09
Add PCRE support

### LINKED ISSUE #1137: Support for pcre2 search pattern?
Any chance to add pcre2 support? such as look-ahead and look-behind.

See https://www.regular-expressions.info/lookaround.html

koutcher:
Closes #1137

### DIFF
--- a/src/search.c
+++ b/src/search.c
@@ -67,13 +67,15 @@ setup_and_find_next(struct view *view, enum request request)
 	    && !utf8_string_contains_uppercase(view->env->search))
 		regex_flags |= REG_ICASE;
 
-	if (view->regex) {
-		regfree(view->regex);
-		*view->grep = 0;
-	} else {
+	if (!view->regex) {
 		view->regex = calloc(1, sizeof(*view->regex));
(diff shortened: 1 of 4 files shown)


src/search.c
```diff
@@ -67,15 +67,17 @@ setup_and_find_next(struct view *view, enum request request)
...
-	} else {
+	if (!view->regex) {
 		view->regex = calloc(1, sizeof(*view->regex));
 		if (!view->regex)
 			return ERROR_OUT_OF_MEMORY;
 	}
 
+	if (*view->grep) {
+		regfree(view->regex);
+		*view->grep = 0;
```
jonas, line 78: Yes, it is more readable to decouple the allocation from regular housekeeping needed when performing a new search.

jonas (approved):
LGTM

src/search.c
koutcher: Yes and also, maybe I should have mentioned it, PCRE has a slight tendency to segfault when using `regfree()` on `view->regex` before `regcomp()`.

PR #1144, opened 2021-09-16
Run macOS CI on pushes and pull requests

krobelus:


jonas:
Nice. I'm completely new to GitHub actions. Great idea to build/test on macOS.

If I remember correctly we only build on master pushes because `brew install --HEAD` only supports building from that branch.

jonas (approved):

PR #1147, opened 2021-09-20
Fix stash view when using tig --all or tig revs

### LINKED ISSUE #1146: Stash view shows the reflog when tig run with a branch name or the `--all` option
Trying to show the stash view when I run `tig --all` shows reflog with the stash view options instead of stash. I experience the same behaviour when I open tig with a specific branch name (`tig <branch>`).

When I run `tig` without options, the stash view works as expected.

I have the latest version of tig, 2.5.4, on macOS via Homebrew.

koutcher:
This is a regression from #1030 due to a wrong handling of --all,
--branches, --remotes and revisions arguments in `git stash list`.
Workaround the issue by filtering them from %(revargs).

Fixes #1146

### DIFF
--- a/src/stash.c
+++ b/src/stash.c
@@ -21,15 +21,35 @@ static enum status_code
 stash_open(struct view *view, enum open_flags flags)
 {
 	static const char *stash_argv[] = { "git", "stash", "list",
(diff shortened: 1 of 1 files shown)


jonas (approved):
Looks good.

PR #1155, opened 2021-10-06
Pass command line args through to diff-files in status view

### LINKED ISSUE #1152: Status takes a long time to load with many submodules
The repository I'm using has lots of submodules. The normal `git status` command takes a long time to run, but if I set `diff.ignoreSubmodules = all` in my git config it runs quickly. This doesn't work with `tig` though because `tig` uses `git diff-files` which doesn't use the value of `diff.ignoreSubmodules = all`.

I can reproduce this manually.

* If I try running `git diff-files -z` it takes a long time.
* If I run `git diff-files --ignore-submodules -z` it is quite a lot faster.

Unfortunately there isn't any option for me to add the `--ignore-submodules` argument to `tig` as far as I can tell.

koutcher:
This makes it possible to use `tig --ignore-submodules` to speed up
the status view for a repo with many submodules.

Closes #1152

### DIFF
--- a/include/tig/git.h
+++ b/include/tig/git.h
@@ -39,7 +39,7 @@
 	"git", "diff-index", (output_arg), "--diff-filter=ACDMRTXB", "-C", "--cached", "HEAD", "--", NULL
 
 #define GIT_DIFF_UNSTAGED_FILES(output_arg) \
-	"git", "diff-files", (output_arg), NULL
+	"git", "diff-files", "%(cmdlineargs)", (output_arg), NULL
 
 #define GIT_DIFF_BLAME(encoding_arg, context_arg, space_arg, new_name) \
 	GIT_DIFF_UNSTAGED(encoding_arg, context_arg, space_arg, "", new_name)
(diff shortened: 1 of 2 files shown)


krobelus:
Looks good. `%(cmdlineargs)` is passed to both log and diff commands,
so it looks like it is correct to forward it to diff-files.

Adding it only for GIT_DIFF_UNSTAGED_FILES and not for GIT_DIFF_STAGED_FILES
is a bit inconsistent but seems good enough since it doesn't matter for the
submodule issue.
This means that with --ignore-submodules, Enter in the status view no longer
does anything on staged submodules. We could sidestep that by not showing
staged submodules (by adding %(cmdlineargs) to GIT_DIFF_STAGED_FILES too).
That would allow to filter staged hunks with `tig -I boilerplate` etc.
Probably absolutely not important.

If the only changes in the worktree are unstaged submodules, `tig --ignore-submodules`
will still show the "Unstaged Changes" in the main view,
but it's empty. Something similar already happens with commands like
`tig -I ''` that filter away every single hunk via `%(cmdlineargs)`.

koutcher:
Thanks for the feedback.

PR #1158, opened 2021-10-26
Refactor: break up nested ternaries into multiple lines

krobelus:
We have some places where multiple ternary operators are nested on a
single line.  Break them up into multiple lines. I think this is much
easier to read because the indentation shows the nesting structure.
No functional change here.

Reading through format_append_arg(), I noticed that "echo %%(commit"
with the missing closing parenthesis is not supported. Probably
not worth spending too much time on that.

### DIFF
--- a/src/argv.c
+++ b/src/argv.c
@@ -337,8 +337,24 @@ format_append_arg(struct format_context *format, const char ***dst_argv, const c
 		const char *var = strstr(arg, "%(");
 		const char *esc = var > arg && *(var - 1) == '%' ? var - 1 : NULL;
 		const char *closing = var ? strchr(var, ')') : NULL;
-		const char *next = esc > arg ? esc : closing ? closing + 1 : NULL;
-		const int len = var && !esc ? var - arg : esc > arg ? esc - arg : next ? next - ++arg : strlen(arg);
+		const char *next;
+		int len;
+
+		if (esc > arg)
+			next = esc;
+		else if (closing)
+			next = closing + 1;
+		else
+			next = NULL;
+
+		if (var && !esc)
+			len = var - arg;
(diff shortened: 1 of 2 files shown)


koutcher:
Not much time needed, I think this should do the job:
```diff
-               if (var && !closing)
+               if (var && !closing && !esc)
```
When doing 6134336, I took it as a challenge to do it in one line only, so I'll have hard time to let it go :-), but, agreed, as we have no plans to enter IOCCC with Tig, we should prioritize the readability.

krobelus:
> if (var && !closing && !esc)

Problem is that `%%(foo` will become `%%(foo`, so there is no way to express `%(foo`.
So the rule should probably be: `%%(` becomes `%(`, and elsewhere `%` is not special.
That would mean `%%%(` becomes `%%(` which feels weird. Also it might be legit to want
`%%(branch)` to output `%some-branch`, which we don't support.
Anyway it's really not important, these are mostly theoretical issues.
But if we make more changes to the expansion code, we should probably reduce the number of special cases.

> I took it as a challenge to do it in one line only

Haha yeah, I was mildly surprised when I came across this while trying to understand the formatting logic

koutcher:
You're right, I missed that point. This is better but as you said, we cannot get `%some-branch`. As in Tig context there is very little reason to want `%(` without closing `)` we may leave it as it is today.

```patch
-               if (var && !closing)
-                       return false;
+               if (var && !closing) {
+                       if (!esc)
+                               return false;
+                       else
+                               ++arg;
+               }
```

krobelus:
Ideally we can interpret each `%%` as `%` (like printf). This has a fairly low risk of breaking scripts.

PR #1159, opened 2021-10-26
Fix null dereferences on unset %(fileargs)

krobelus:


koutcher:
What about:
```patch
diff --git a/src/argv.c b/src/argv.c
index 5ba7fc1..2e207a0 100644
--- a/src/argv.c
+++ b/src/argv.c
@@ -439,7 +439,7 @@ argv_format(struct argv_env *argv_env, const char ***dst_argv, const char *src_a
                const char *arg = src_argv[argc];

                if (!strcmp(arg, "%(fileargs)")) {
-                       if (file_filter && !argv_append_array(dst_argv, opt_file_args))
+                       if (file_filter && !format_append_argv(&format, dst_argv, opt_file_args))
                                break;

                } else if (!strcmp(arg, DIFF_ARGS)) {
```
It fails the part of the test you modified but only to be consistent with the original test result (and the comment inside the test).

test/regressions/github-1136-test
```diff
@@ -13,21 +13,54 @@ test_case bang-cmdlineargs-doesnt-crash \
...
 	' <<EOF
+On branch master
+Changes to be committed:
+  (no files)
+Changes not staged for commit:
+  (no files)
+Untracked files:
+  (no files)
 
+[status] Nothing to update                                                  100%
```
koutcher, line 24: This is not consistent with the comment "This runs an empty command, hence the empty pager".

test/regressions/github-1136-test
```diff
@@ -13,21 +13,54 @@ test_case bang-cmdlineargs-doesnt-crash \
...
+Changes not staged for commit:
+  (no files)
+Untracked files:
+  (no files)
 
-
-
-
-[pager]                                                                       0%
+[status] Nothing to update                                                  100%
```
koutcher, line 57: Same here.

krobelus:
Sorry for the delay.

> What about:
> \-                       if (file_filter && !argv_append_array(dst_argv, opt_file_args))
> \+                       if (file_filter && !format_append_argv(&format, dst_argv, opt_file_args))

One possible surprise here is that format_append_argv() treats pathspecs as Tig format strings, for example in

	tig -- './%(fileargs)'
	tig: Failed to format main arguments

(We do interpret others like mainargs and cmdlineargs as format strings. I don't know if that's used anywhere.)

My original fix turned `"%(fileargs)"` into `{NULL}` while yours turns it into `{"", NULL}`. I think `{""}` works fine because argv_parse_rev_flag() will filter away the empty string, and for %(fileargs) it doesn't matter since the empty pathspec matches all files. You're right that my test changes were weird but it's also a weird edge case.

Probably the current state is fine. I'm sure there is a way to simplify this string handling code next time we touch it.

(BTW there's another crash on `:toggle file-args`. Not very important of course)

koutcher:
Thanks Johannes, you're right. I don't like it either, even considering the low probability. Let's revert to using argv_append() but with a check for an empty argument. It's simpler than your original proposal, still relying on `{"", NULL}` being acceptable but also still working with the original test case.
```patch
diff --git a/src/argv.c b/src/argv.c
index 92403f4..b259cfc 100644
--- a/src/argv.c
+++ b/src/argv.c
@@ -441,7 +441,9 @@ argv_format(struct argv_env *argv_env, const char ***dst_argv, const char *src_a
                const char *arg = src_argv[argc];

                if (!strcmp(arg, "%(fileargs)")) {
-                       if ((flags & argv_flag_file_filter) && !format_append_argv(&format, dst_argv, opt_file_args))
+                       if ((flags & argv_flag_file_filter) &&
+                           ((!opt_file_args && !argv_append(dst_argv, "")) ||
+                            (!argv_append_array(dst_argv, opt_file_args))))
                                break;

                } else if (!strcmp(arg, DIFF_ARGS)) {
@@ -466,7 +468,9 @@ argv_format(struct argv_env *argv_env, const char ***dst_argv, const char *src_a

                } else if (!strcmp(arg, "%(revargs)") ||
                           ((flags & argv_flag_first) && !strcmp(arg, "%(commit)"))) {
-                       if ((flags & argv_flag_rev_filter) && !format_append_argv(&format, dst_argv, opt_rev_args))
+                       if ((flags & argv_flag_rev_filter) &&
+                           ((!opt_rev_args && !argv_append(dst_argv, "")) ||
+                            (!argv_append_array(dst_argv, opt_rev_args))))
                                break;

                } else if (!format_append_arg(&format, dst_argv, arg)) {
```

For `:toggle file-args` we could do:
```patch
diff --git a/src/prompt.c b/src/prompt.c
index 13d6ae7..8a41175 100644
--- a/src/prompt.c
+++ b/src/prompt.c
@@ -798,8 +798,10 @@ prompt_toggle_option(struct view *view, const char *argv[], const char *prefix,
                int i;

                if (argv_size(argv) <= 2) {
-                       argv_free(*opt);
-                       (*opt)[0] = NULL;
+                       if (*opt) {
+                               argv_free(*opt);
+                               (*opt)[0] = NULL;
+                       }
                        return SUCCESS;
                }
```

krobelus:
Both sound good to me, though a ternary operator might be more intuitive here:

	!(opt_file_args ? argv_append_array(...) : argv_append(..., ""))

Issue #1160, opened 2021-10-27
Hiding refs by pattern (to reduce git-branchless spam)?


nyanpasu64:
Installing git-branchless has resulted in it labeling large numbers of commits with meaningless refs which clutter the tig view. Is it possible to hide certain refs from tig by a regex/glob pattern?

![Screenshot_20211026_180119](https://user-images.githubusercontent.com/913957/138982527-edd72f22-1af9-41cb-8e77-1078afcda4d6.png)

I looked at https://jonas.github.io/tig/doc/tigrc.5.html#_variables, and it seems the closest you can get is `hide:`, but you can't hide branches and unknown refs separately, nor can you hide `refs/stash` and `refs/branchless` separately if you want.

koutcher:
We could also introduce a new reference category `other` to allow the use of `set reference-format = hide:other` to hide these references which aren't heads, tags, branches...

PR #1161, opened 2021-10-28
Remove quotes from TIG_LS_REMOTE to allow inline shell commands

krobelus:
The environment variable TIG_LS_REMOTE can be used to select the
references that Tig will display.

When using a command like TIG_LS_REMOTE='sh -c "a | b"' tig, we pass an
argv-array with 3 elements to execvpe.  The third argument is passed
with quotes, which means that the shell tries to run a command that
is literally called 'a | b'.

Fix this by removing quotes. We already split arguments with a
shell-like syntax, so this only improves compatibility.

This makes it easier filter away unwanted refs (#1160) without creating
a script TIG_LS_REMOTE.

### DIFF
--- a/src/refdb.c
+++ b/src/refdb.c
@@ -325,7 +325,7 @@ reload_refs(bool force)
 		int argc = 0;
 
 		string_ncopy(ls_remote_cmd, env, strlen(env));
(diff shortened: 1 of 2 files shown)


koutcher (approved):
LGTM

PR #1162, opened 2021-11-11
Fix navigation between chunks after splitting chunk

### LINKED ISSUE #1101: Navigating between chunks doesn't work after splitting chunk
When in status view, pressing "@" navigates between chunks. If I then choose to split the current chunk with "\\", it is no longer possible to navigate between the chunks with "@", the alignment seems to be off: the cursor no longer ends up at the first line of each chunk. My guess is that the underlying git state and the tig view is somehow misaligned. Navigating line by line (with "j" and "k") still works, but jumping between whole chunks is unusable after a split.

koutcher:
Also fix the color of the first split chunk.

Closes #1101

### DIFF
--- a/src/stage.c
+++ b/src/stage.c
@@ -25,6 +25,7 @@
 #include "tig/status.h"
 #include "tig/main.h"
 #include "tig/stage.h"
+#include "tig/search.h"
 
 static struct status stage_status;
 static enum line_type stage_line_type;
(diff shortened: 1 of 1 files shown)


src/stage.c
```diff
@@ -607,6 +610,7 @@ stage_request(struct view *view, enum request request, struct line *line)
 			return REQ_NONE;
 		}
 		stage_split_chunk(view, line);
+		reset_search(view);
```
krobelus, line 613: Looks obviously correct. I guess we could do this on successful splits only, but I don't see a good reason to favor one over the other

```diff
diff --git a/src/stage.c b/src/stage.c
index 1b0a92e2..0e869381 100644
--- a/src/stage.c
+++ b/src/stage.c
@@ -404,6 +404,7 @@ stage_split_chunk(struct view *view, struct line *chunk_start)
 	if (chunks) {
 		stage_insert_chunk(view, &header, chunk_start, NULL, NULL);
 		redraw_view(view);
+		reset_search(view);
 		report("Split the chunk in %d", chunks + 1);
 	} else {
 		report("The chunk cannot be split");
```

PR #1169, opened 2021-11-24
Update distclean rule to preserve tarball contents

### LINKED ISSUE #1167: "make distclean" removes too much
The `distclean` target in the release tarball removes files that are part of the tarball (and thus the released source). In other words, if I unpack a tarball and run `make distclean`, I lose e.g. `configure`, `doc/tig.1`, etc. This is especially problematic when trying to build a Debian package: the package builder will run `make distclean` first. It knows how to run `autogen.sh` to get `configure` back, but other files such as the manual pages are lost.

In my understanding, the `distclean` target should return the tree to the state of the distributed source code. If the distributed source contains a `configure` script and (generated) man pages, these should *not* be removed (I would suggest a `pristine` target for that).

koutcher:
... and replace it with new realclean rule.

Closes #1167

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,15 +176,18 @@ uninstall:
 		$(QUIET_UNINSTALL_EACH)tools/uninstall.sh "$(DESTDIR)$(docdir)/tig/$(doc)";)
 
 clean: clean-test clean-coverage
-	$(Q)$(RM) -r $(TARNAME) *.spec tig-*.tar.gz tig-*.tar.gz.sha256 .deps _book node_modules
+	$(Q)$(RM) -r $(TARNAME) tig-*.tar.gz tig-*.tar.gz.sha256 .deps _book node_modules
 	$(Q)$(RM) -r $(compdb_dir) compile_commands.json
 	$(Q)$(RM) $(EXE) $(TOOLS) $(OBJS) core doc/*.xml src/builtin-config.c
 	$(Q)$(RM) $(OBJS:%.o=%.gcda) $(OBJS:%.o=%.gcno)
 
 distclean: clean
 	$(RM) -r doc/manual.html-chunked autom4te.cache
-	$(RM) doc/*.toc $(ALLDOC) aclocal.m4 configure
-	$(RM) config.h config.log config.make config.status config.h.in
+	$(RM) doc/*.toc configure~
+	$(RM) config.h config.log config.make config.status config.h.in~
+
+veryclean: distclean
+	$(RM) tig.spec $(ALLDOC) aclocal.m4 configure config.h.in
 
 spell-check:
 	for file in $(TXTDOC) src/tig.c; do \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -176,15 +176,18 @@ uninstall:
...
 	$(Q)$(RM) $(OBJS:%.o=%.gcda) $(OBJS:%.o=%.gcno)
 
 distclean: clean
 	$(RM) -r doc/manual.html-chunked autom4te.cache
-	$(RM) doc/*.toc $(ALLDOC) aclocal.m4 configure
-	$(RM) config.h config.log config.make config.status config.h.in
+	$(RM) doc/*.toc configure~
+	$(RM) config.h config.log config.make config.status config.h.in~
+
+realclean: distclean
```
krobelus, line 189: cc @sbakker

I also thought that `make clean` deletes all generated files, while `make distclean` preserves things like documentation and configure scripts.

In Tig it's the other way round, which is weird. However, Git does the same, so we should first understand why to see if it makes sense for us to deviate.

AFACT this behavior was originally introduced to Git in https://github.com/git/git/commit/922aa807f5fee97fcb2ccd1d1db63b59bdd02128
this is the log message:

	Do not remove distributed configure script

	Before this patch the clean target has removed the
	configure script that comes with Git tar file.
	That made compiling Git for different architectures
	inconvenient.
	This patch excludes configure from the files to be
	deleted by 'make clean' and adds new target 'distclean'
	to preserve old functionality.

(if we stick to a separate target, maybe clean-all is a more familiar/discoverable name)

koutcher:
What makes things a bit more complicated with Tig is the two-stage delivery process which adds generated artefacts that may not be reproducible to the delivery branch before using it to build the tarball. The current `distclean` behaviour is consistent from the point view of the master branch but not when you get a tarball. As for the name, I didn't find enough reference to `pristine` proposed by @sbakker so I preferred `realclean` which is mentioned in Make documentation, though often synonym to `distclean`. With `clean-all` there is a risk of confusion with the classic `make clean all`.

Makefile
sbakker: I'm not stuck on the `pristine` name, it's just something I saw occasionally (possibly along with `mrproper` in the Linux kernel source). The [Standard Targets section(https://www.gnu.org/software/make/manual/html_node/Standard-Targets.html#Standard-Targets) in the [GNU make manual](https://www.gnu.org/software/make/manual/) defines `clean` and `distclean`:
  * `clean` - delete files created by the Makefile during the build (compilation) process (`.o`, `.a`, `.so`, etc.), but not configuration files.
  * `distclean` - delete all files that are created by configuring or building the program (so also `config.status`, but not the `configure` script). It should return the state of the directory to that of the unpacked tarball.
 There's also a `maintainer-clean`, but even that one is not supposed to delete the `configure` script. So `veryclean` is probably the best choice here.

koutcher: `veryclean` would be good for me.

PR #1172, opened 2021-11-26
Add reference types stash and other

### LINKED ISSUE #1160: Hiding refs by pattern (to reduce git-branchless spam)?

koutcher:
This makes it possible to filter refs created by tools such as git-branchless
with `set reference-format = hide:other`.

Closes #1160

### DIFF
--- a/doc/tigrc.5.adoc
+++ b/doc/tigrc.5.adoc
@@ -185,12 +185,14 @@ The following variables can be set:
 	Prefix with `hide:` to not show that reference type, e.g. `hide:remote`.
 	Supported reference types are:
 	 - head			: The current HEAD.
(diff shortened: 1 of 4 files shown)


dgmstuart:
Thanks so much @koutcher - this has saved my Tig experience 🙏

PR #1173, opened 2021-11-28
Allow to toggle revision filtering

krobelus:


koutcher:
Thanks, I found myself several times having to restart tig without revision argument when playing with the blame view, this will be really useful. 
There are more variables that do not trig a refresh as they should. I remembered checking that a while ago but I never got time to test it well enough to make a PR. I rebased it and pushed it at https://github.com/koutcher/tig/commits/wip-option-refresh if you want to have a look.

PR #1178, opened 2022-01-21
Fix `:goto <tag>`

koutcher:
Dereference goto argument to make sure we try to reach the commit and
not the tag object for annotated tags.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -333,7 +333,8 @@ goto_id(struct view *view, const char *expr, bool from_start, bool save_search)
 		report("Jumping to ID is not supported by the %s view", view->name);
 		return;
 	} else {
-		char *rev = argv_format_arg(view->env, expr);
+		char tmp[SIZEOF_STR];
+		char *rev = (snprintf(tmp, SIZEOF_STR, "%s^{}", expr), argv_format_arg(view->env, tmp));
 		const char *rev_parse_argv[] = {
 			"git", "rev-parse", "--revs-only", rev, NULL
 		};


src/prompt.c
```diff
@@ -959,10 +959,14 @@ run_prompt_command(struct view *view, const char *argv[])
...
-		if (!argv[1] || !strlen(argv[1]))
+		if (!argv[1] || !strlen(argv[1])) {
 			report("goto requires an argument");
-		else
-			goto_id(view, argv[1], true, true);
+		} else {
+			char dest[SIZEOF_STR];
+
+			snprintf(dest, SIZEOF_STR, "%s^{}", argv[1]);
+			goto_id(view, dest, true, true);
```
krobelus, line 968: LGTM

koutcher: Thanks for the review Johannes. As one can also just do `:<sha1>` at the prompt, I'll amend the PR to move the code directly in goto_id().

src/view.c
```diff
@@ -333,7 +333,8 @@ goto_id(struct view *view, const char *expr, bool from_start, bool save_search)
 		report("Jumping to ID is not supported by the %s view", view->name);
 		return;
 	} else {
-		char *rev = argv_format_arg(view->env, expr);
+		char tmp[SIZEOF_STR];
+		char *rev = snprintf(tmp, SIZEOF_STR, "%s^{}", expr) ? argv_format_arg(view->env, tmp) : NULL;
```
krobelus, line 337: Looks good although the ? is weird (I don't think we'll ever get 0). Maybe ignore the return value explicitly using the comma operator:

```suggestion
		char *rev = (snprintf(tmp, SIZEOF_STR, "%s^{}", expr), argv_format_arg(view->env, tmp));
```

(The upcoming Git 2.35 has a "Weather balloon to break people with compilers that do not support C99.". Maybe we should try that too, then this will be more obvious)

koutcher: You're right, the test should have been at least `... > 0` but it doesn't make much sense here anyway. I was looking for other uses of the comma operator in Tig and found one strange one in `ui.c:303`. Looks like a typo for me, what do you think ?

As tig still works well with old Git versions, I think it is too early for us to think about dropping old compilers.

krobelus:
> I was looking for other uses of the comma operator in Tig and found one strange one in `ui.c:303`

Yep that definitely looks like a typo.

I didn't think of using clang-query before, it's so easy:

	clang-query src/*.[ch] -c 'match binaryOperator(hasOperatorName(","))'

If the comma operator is too subtle here, we can also be more explicit about ignoring the return value with `snprintf()*0 +argv_format_view()`

PR #1190, opened 2022-05-26
Change the blame view to use porcelain

### LINKED ISSUE #978: Display only the specified range when blaming
`git blame` supports `-L` option with which we can supply a range of lines to blame, and git outputs the blame for **only** those lines. This comes in handy when invoking from within an editor, say Vim. For example, to view the blame from lines 2000 to 2020, I could say `:!git blame -L2000,2020 -- %`.

Two issues with `tig`:

* `tig` does accept the same option. However, `tig` outputs **all** lines in the file. This means, for the example above, even though I'm interested in the 20 lines towards the bottom, `tig` shows the whole file, and I've to scroll 2000 lines for the view I'm interested in. Any chance this can supported?
* To `tig`, I've to specify it as `-L2000,2020`; it seems to be confused if I say `-L 2000,2020`, with white space after `-L`.

### LINKED ISSUE #1189: blame does not honor textconv (but diff does)
Unfortunately, tig blame does not use the textconv configuration like git blame or git gui blame do for my repository. On the plus side, tig does use textconv configuration when showing patches/diffs. Ideally, tig blame would use textconv if the repository is configured for it.

koutcher:
Fixes #978
Fixes #1189

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -15,6 +15,7 @@ Bug fixes:
  - Fix cursor behaviour during staging. (#842, #1028)
  - Fix navigation in split tree view.
  - Enable textconv in the stage view.
+ - Enable textconv in the blame view and fix blame -L. (#978, #1189)
 
 tig-2.5.5
 ---------
--- a/src/blame.c
+++ b/src/blame.c
@@ -49,9 +49,8 @@ struct blame {
 
 struct blame_state {
 	struct blame_commit *commit;
+	struct blame_header header;
 	char author[SIZEOF_STR];
-	int blamed;
-	bool done_reading;
 	bool auto_filename_display;
 	const char *filename;
 	/* The history state for the current view is cached in the view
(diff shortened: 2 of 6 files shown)


NEWS.adoc
```diff
@@ -15,6 +15,7 @@ Bug fixes:
  - Fix cursor behaviour during staging. (#842, #1028)
  - Fix navigation in split tree view.
  - Enable textconv in the stage view.
+ - Change the blame view to use porcelain. (#978, #1189)
```
krobelus, line 18: I'm not sure if porcelain vs. incremental is relevant for users. I'd probably say something like

The blame view now honors the textconv option and blame flags like -L.

src/blame.c
```diff
@@ -210,77 +207,27 @@ get_blame_commit(struct view *view, const char *id)
...
-
 	return commit;
 }
 
-static bool
-blame_read_file(struct view *view, struct buffer *buf, struct blame_state *state)
-{
-	if (!buf) {
-		const char *blame_argv[] = {
-			"git", "blame", encoding_arg, "%(blameargs)", "--incremental",
```
krobelus, line 242: The implementation looks solid. I'm wondering though, is it really necessary to drop `--incremental` to support `-L` and `--textconv`?

Probably not a huge loss although incremental mode can be nice.
In theory `git blame --incremental -C +123` could try to find the commit that introduced line 123 first, but that's not what happens today.

krobelus:
I just realized this introduced a regression: we no longer jump to the blamed line

To reproduce run

    tig show 2280734104fd362a1188b0678109e01cdb9322cf +46

and type `b`.

koutcher:
Good catch, the goto_lineno has gone with the bathwater, I'll put it back.

PR #1211, opened 2022-07-06
Fix reading the value of `status.showUntrackedFiles`

storoj:
It seems like `git config --list` prints everything in lowercase. That results in `tig` not recognising the value of `status.showUntrackedFiles` since it's being printed as `status.showuntrackedfiles`.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -1484,7 +1484,7 @@ read_repo_config_option(char *name, size_t namelen, char *value, size_t valuelen
 	else if (!strcmp(name, "diff.noprefix"))
 		parse_bool(&opt_diff_noprefix, value);
 
-	else if (!strcmp(name, "status.showUntrackedFiles"))
+	else if (!strcmp(name, "status.showuntrackedfiles"))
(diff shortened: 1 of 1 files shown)


src/options.c
```diff
@@ -1484,7 +1484,7 @@ read_repo_config_option(char *name, size_t namelen, char *value, size_t valuelen
 	else if (!strcmp(name, "diff.noprefix"))
 		parse_bool(&opt_diff_noprefix, value);
 
-	else if (!strcmp(name, "status.showUntrackedFiles"))
+	else if (!strcmp(name, "status.showuntrackedfiles"))
```
krobelus, line 1487: nice fix, thanks!
If possible, could you update the commit subject (the first line in the commit message) to mention `status.showUntrackedFiles` like the PR subject does? The commit subject will appear in the release notes (via git-shortlog).

storoj: Done!

koutcher: Thanks.

Issue #1215, opened 2022-07-15
~/.local/share/tig/history may not be used since version 2.5.6


yan12125:
Since 2.5.6 (d308b74840e8b7ce8e5d00798bf578d3c92f4d92, particularly), if `$XDG_DATA_HOME` is valid, tig runs `mkdir(dirname(path), 0777);`. However, dirname() may modify the string pointed to by path [1]. If that happens, path no longer points to the desired history file.

For example, on Arch LInux, dirname() changes path from
```c
"/home/yen/.local/share/tig/history\0"
```
to
```c
"/home/yen/.local/share/tig\0history\0"
```
The following open() fails as path is now a directory, and the fallback ~/.tig_history is used.

By the way, is it a good idea to use permission 0777?

[1] https://pubs.opengroup.org/onlinepubs/9699919799/functions/dirname.html

yan12125:
Figured out a fix: https://github.com/jonas/tig/pull/1217

koutcher:
@yan12125 , thanks for reporting and fixing the issue. To my dismay, macOS version of dirname() does not behave the same as Linux version, so I missed that. 0777 is further restricted by the user's umask, I don't think we have a reason to be more restrictive.

yan12125:
> 0777 is further restricted by the user's umask, I don't think we have a reason to be more restrictive.

Got it!

PR #1217, opened 2022-07-16
Avoid incorrect history path when $XDG_DATA_HOME is valid

yan12125:
Fixes https://github.com/jonas/tig/issues/1215

### DIFF
--- a/src/prompt.c
+++ b/src/prompt.c
@@ -11,6 +11,7 @@
  * GNU General Public License for more details.
  */
 
+#include <string.h>
 #include "tig/tig.h"
 #include "tig/repo.h"
 #include "tig/view.h"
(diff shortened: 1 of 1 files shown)


src/prompt.c
```diff
@@ -538,8 +539,11 @@ prompt_histfile(void)
 			die("Failed to expand $HOME");
 	} else if (!string_format(path, "%s/tig/history", xdg_data_home))
 		die("Failed to expand $XDG_DATA_HOME");
-	else
-		mkdir(dirname(path), 0777);
+	else {
+		static char path_copy[SIZEOF_STR] = "";
```
koutcher, line 543: `path_copy` won't be used outside of prompt_histfile(), so we don't need to make it static. 
```
char path_copy[SIZEOF_STR];
```
should be enough here, don't you think ?

yan12125: That looks better. I don't remember why I added `static` :joy:

PR #1223, opened 2022-07-21
Autoscroll in pager

psprint:
A second attempt (after #1139) to add two features:
- autoscroll in pager,
- regex matching lines colorization in pager.

In this attempt I've did what has been recommended to me previously:
- autoscroll is off by default so that `tig`  resembles `less` (`set pager-autoscroll = off`),,
- the regex used by the line colorization can be also customized via an option `color-pager-regex`.

I think that running `make` in pager is a good use case and it could be better supported.

### DIFF
--- a/src/view.c
+++ b/src/view.c
@@ -621,6 +621,7 @@ begin_update(struct view *view, const char *dir, const char **argv, enum open_fl
 bool
 update_view(struct view *view)
 {
+        bool should_autoscroll = false;
 	/* Clear the view and redraw everything since the tree sorting
 	 * might have rearranged things. */
 	bool redraw = view->lines == 0;
@@ -652,11 +653,19 @@ update_view(struct view *view)
 			return false;
 		}
 
+		if ((view->pos.offset + view->height + 1) == view->lines)
+		    should_autoscroll = true;
+
 		if (!view->ops->read(view, &line, false)) {
 			report("Allocation failure");
 			end_update(view, true);
 			return false;
 		}
+
+		/* Autoscroll is available only in pager */
+		if (should_autoscroll && opt_pager_autoscroll &&
+                    !strcmp(view->name, "pager"))
+		        do_scroll_view(view, 2);
 	}
 
 	if (io_error(view->pipe)) {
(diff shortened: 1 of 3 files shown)


psprint:
@kumanna ping. I've worked on the patch and took all suggestions into account. I wonder what chances have this PR to merge?

psprint:
@koutcher ping :)

psprint:
@koutcher @krobelus merge?

psprint:
@koutcher @krobelus ping :wave:

src/view.c
```diff
@@ -652,11 +653,19 @@ update_view(struct view *view)
 			return false;
 		}
 
+		if ((view->pos.offset + view->height + 1) == view->lines)
```
krobelus, line 656: autoscroll seems like an interesting feature. I don't use the pager view but I can imagine autoscroll being useful.
I think this commit should be submitted in a separate PR, without the highlighting changes.

There seems to be an off-by-one error. When I type `:!git log` it only scrolls until the last-but-one line; the last line is not visible.

Instead of an option, we could also implement less's F command.

src/diff.c
```diff
@@ -302,12 +302,38 @@ diff_common_highlight(struct view *view, const char *text, enum line_type type)
...
+            /* Parse all compiler message lines, ie. such as:
+             *
+             * {file}:{line}:{col}: (note|warning|error):{message} 
+             *
+             */
+            if (opt_color_pager_regex != NULL && *opt_color_pager_regex)
+                regex_txt = opt_color_pager_regex;
+            else
+                /* A reasonable default - match GCC errors */
+                regex_txt = "([^:]+):([0-9]+):(|([0-9]+):)[ \t]+(note|warning|error): .*$";
```
krobelus, line 327: again, this shouldn't be coded into Tig (sources or tigrc) because Tig isn't typically used to display compiler output.

src/diff.c
```diff
@@ -302,12 +302,38 @@ diff_common_highlight(struct view *view, const char *text, enum line_type type)
...
+            regex_err = regcomp(&regex, regex_txt, regex_flags);
+
+            if (!regex_err)
+                regex_err = regexec(&regex, data, 8, pmatch, 0);
+
+            if (!regex_err) 
+                type = LINE_COLOR_MATCH;
+            regfree(&regex);
```
krobelus, line 335: if regexec returns an error we are freeing an uninitialized regex_t here, which is bad.

(also you have trailing whitespace here)

src/view.c
psprint: @krobelus: I've updated the PR to cover only autoscroll feature and fixed the off by 1 error. I agree that autoscroll is useful, I'm using it all the time.

koutcher:
Thanks. It needed a bit of rework to have things done at the proper place.

psprint:
Thanks for merging :)

PR #1231, opened 2022-08-31
A new substitution special string %(sh {shell-command}) that substitutes external command's output

psprint:
On gitter, in a [discussion](https://gitter.im/jonas/tig?at=62ae474bef5ee44882b4e993) there has been an idea from user @krobelus:

```
<krobelus> I don't thikn this is possible, we should add that feature
10:17 maybe like :goto %sh{git rev-list %(commit) | head -1
```

– it was an idea of a %(…)-like substitution string that would substitute the given external command's output. I have implemented the idea in this PR, as `%(sh …)` special string, similar to `%(prompt …)`.

This PR implements a new substitution special string `%(sh {shell-command})` that runs an external command, captures its output and substitutes it in place of the `%`-string. It can be used as follows:

```
bind generic T :echo '%(sh printf "Home dir is $HOME")'
```

As it can be seen, the command is run via `sh -c '{command}'`, so environment variable substitutions can be used.

The implementation works OK with limitation inherited from `%(prompt )` – in the argument there shouldn't be any other substitutions like `%(commit)`, because they will not be substituted. Any ideas were in the code to implement this? It would be needed for e.g.:

```
bind generic T :goto '%(sh git rev-list %(commit)~..|head -1)'
```

### DIFF
--- a/src/argv.c
+++ b/src/argv.c
@@ -315,6 +315,29 @@ format_expand_arg(struct format_context *format, const char *name, const char *e
 			return false;
 		return string_format_from(format->buf, &format->bufpos, "%s", value);
 	}
+	if (!prefixcmp(name, "%(sh")) {
+		char cbuf[SIZEOF_STR];
+		const char *c="";
+		char value[SIZEOF_STR]={0};
+		const char *cstart = name + STRING_SIZE("%(sh");
+		const int clen = end - cstart - 1;
+		int size;
+		FILE *cp;
+		if (end && clen > 0 && string_format(cbuf, "%.*s", clen, cstart)) {
+			c=cbuf;
+			while (isspace(*c))
+				c++;
+		}
+		if (!*c||strlen(c)==8)
+			return false;
+		cp=popen(c,"r");
+		size=fread(value,1,SIZEOF_STR-1,cp);
+		if (value[0]==0 || size==0)
+			return false;
+		if (value[size-1]=='\n')
+			value[size-1]='\0';
+		return string_format_from(format->buf, &format->bufpos, "%s", value);
+    	}
 
 	for (i = 0; i < format->vars_size; i++) {
 		if (string_enum_compare(name, vars[i].name, vars[i].namelen))
--- a/src/argv.c
+++ b/src/argv.c
@@ -300,7 +300,7 @@ format_expand_arg(struct format_context *format, const char *name, const char *e
 		const char *value;
 		const char *msgstart = name + STRING_SIZE("%(prompt");
 		const int msglen = end - msgstart - 1;
-
+		char *tmp;
 		if (end && msglen > 0 && string_format(msgbuf, "%.*s", msglen, msgstart)) {
 			const char *msg = msgbuf;
 
@@ -310,7 +310,9 @@ format_expand_arg(struct format_context *format, const char *name, const char *e
 				prompt = msg;
 		}
 
-		value = read_prompt(prompt);
+		tmp=argv_format_arg(&argv_env, prompt);
+		value = read_prompt(tmp);
+		free(tmp);
 		if (value == NULL)
 			return false;
 		return string_format_from(format->buf, &format->bufpos, "%s", value);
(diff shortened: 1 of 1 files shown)


krobelus:
I agree the expansion system could use some love.
The current one is inspired by Git's format strings that allow things like `%(refname:strip=3)`.
Obviously Git doesn't need shell expansions but we probably do.

> bind generic T :goto '%(sh git rev-list %(commit)~..|head -1)'

I feel like it's a bad idea to substitute `%(commit)` here.
I'd rather make users write

	bind generic T :goto '%(sh git rev-list "$tig_commit"~..|head -1)'

then the contents of the %(sh) block is pure shell, no other funny syntax.

Another potential problem is that you have single quotes around the shell expansion.
How does that interact with single quotes inside the shell expansion?
They should not affect each other. Not sure if this is the case.

src/argv.c
```diff
@@ -315,6 +315,29 @@ format_expand_arg(struct format_context *format, const char *name, const char *e
...
 	}
+	if (!prefixcmp(name, "%(sh")) {
+		char cbuf[SIZEOF_STR];
+		const char *c="";
+		char value[SIZEOF_STR]={0};
+		const char *cstart = name + STRING_SIZE("%(sh");
+		const int clen = end - cstart - 1;
+		int size;
+		FILE *cp;
+		if (end && clen > 0 && string_format(cbuf, "sh -c '%.*s'", clen, cstart)) {
```
krobelus, line 326: if the command can contain single quotes, this will go wrong.
We should either escape single quotes (replacing them with `'\''`) or,
better, don't spawn two shells here, one is enough.

src/argv.c
```diff
@@ -315,6 +315,29 @@ format_expand_arg(struct format_context *format, const char *name, const char *e
...
+		}
+		if (!*c || strlen(c)==8)
+			return false;
+		cp=popen(c,"r");
+		size=fread(value,1,SIZEOF_STR-1,cp);
+		if (value[0]==0 || size==0)
+			return false;
+		if (value[size-1]=='\n')
+			value[size-1]='\0';
+		return string_format_from(format->buf, &format->bufpos, "%s", value);
```
krobelus, line 339: the entire output becomes one string argument.
This is good for some cases but if we want to compute multiple arguments for a tig command, it becomes awkward.

An `eval` command might remedy that but I'm not sure if this is a good design, see also https://en.wikipedia.org/wiki/Inner-platform_effect

Maybe instead of

	bind generic T :goto '%(sh git rev-list %(commit)~..|head -1)'

we should write

	bind generic T %echo goto "$(git rev-list "$tig_commit"~..|head -1)"

where the % is the magic key for "run this shell command and run the output as tig command"

Something like this seems like a reasonably general approach but I'm sure there are many things I didn't consider.
Sadly I don't have much time these days.

psprint:
Patch 4301605 contains handling of embedded, recursive `%()`-substitutions. Even `%(sh printf %(sh printf $HOME))` is possible, and also the needed `%(sh printf %(commit))`. @krobelus review?I've read your previous review and I'm  taking it into account.

src/argv.c
psprint: Do you mean that it's not possible to embed `%(commit(` and others inside `%(sh …)`? When you write: `the entire output becomes one string argument.`? Or something else? Because I've implemented support for this in second patch 4301605b.

src/argv.c
psprint: Yes, I think  that the subshell can be skipped, it would be concise with the usual ``bind generic {key} !sh -c '…'`` wrapping ↔ no wrapping → no fancy features like env-vars.

psprint:
@koutcher ping

psprint:
@krobelus ping?

psprint:
The patch is precious also because `%(prompt %(commit))` works as expected…

psprint:
@koutcher @krobelus merge?

src/argv.c
```diff
@@ -310,11 +310,39 @@ format_expand_arg(struct format_context *format, const char *name, const char *e
...
+		const int clen = end - cstart - 1;
+		int size;
+		char *tmp;
+		FILE *cp;
+		if (end && clen > 0 && string_format(cbuf, "%.*s", clen, cstart)) {
+			c=cbuf;
+			while (isspace(*c))
+				c++;
+		}
+		if (!*c||strlen(c)==8)
```
krobelus, line 334: what if this fixed-size-buffer overflows?

Issue #1240, opened 2022-10-20
Compiling with ncurses instead of ncursesw breaks alignment/repaint


mqudsi:
Compiling tig (from `master`) with `libncurses-dev` but no `libncursesw5-dev` installed causes tig to experience some alignment issues and repaint problems that do not happen with `libncursesw5-dev` installed/used instead.

Here's how it looks with the following configuration:

```
...
configure: WARNING: The found ncurses library does not support wide-char.
configure: WARNING: This means that tig will not correctly render UTF-8.
...
```

giving 

```
tig version 2.5.7
ncursesw version 6.1.20180127 # this is wrong?
readline version 7.0
PCRE2 version 10.31 2018-02-12
```

https://user-images.githubusercontent.com/606923/197015041-bfb6fbbd-8ed6-4e5f-b38b-51bae9172a97.mp4

And here's how it looks compiled with `libncursesw5-dev` correctly installed instead:

```
tig version 2.5.7
ncursesw version 6.1.20180127
readline version 7.0
PCRE2 version 10.31 2018-02-12
```

https://user-images.githubusercontent.com/606923/197010343-65f7a82e-a993-4c1e-9ab1-7bdc8023f5ef.mp4

Behavior observed under both Kitty and conhost (Windows cmd). Behavior not reproducible in some other terminal emulators.

I don't think UTF-8 is a factor, at least it shouldn't be as I don't think I see any non-ASCII characters in the output on screen at a glance. If tig *is* using Unicode/non-ASCII symbols for display/formatting, then it *shouldn't* if no libncursesw was detected (in which case wide character support would only be used for git messages or file contents containing Unicode characters).

krobelus:
@mqudsi No idea what's the cause of this issue but it's nice that you're also using Tig :)
I use it for all my code archaeology and code review needs.
I wanna work on it some more, add new UI features etc. but the code can be quite dense.

mqudsi:
Hey, hey @krobelus. Nice to bump into you here too - I had to double-check I had the right repository when I ran `tig` and saw your name come up here as well!

I use tig to see relationships between branches and - most importantly - to navigate backwards in `git blame` history. The ergonomics aren't the greatest (I'm annoyed by how paging in the left and right hand panes necessarily has different bindings), but it gets the job done.

WRT to the behavior described in this bug: it's old and goes back to 2018. Not sure if bisecting will help, but if I have some time I can give that a try.

mqudsi:
[...]

PR #1241, opened 2022-10-20
Correctly report which version of libncurses was linked

mqudsi:
Later versions of ncurses (20111030 and up) support NCURSES_WIDECHAR even when compiling with libncurses instead of libncursesw.

Have tig correctly report whether it was linked against libncursesw or libncurses based off the same defines we set in the build script and use to determine which headers to include.

(See #1240 as an example of where `tig --version` incorrectly reports `ncursesw` instead of `ncurses`).

### DIFF
--- a/src/tig.c
+++ b/src/tig.c
@@ -557,7 +557,7 @@ parse_options(int argc, const char *argv[], bool pager_mode)
 				printf("tig version %s\n", TIG_VERSION);
 #ifdef NCURSES_VERSION
 				printf("%s version %s.%d\n",
-#ifdef NCURSES_WIDECHAR
+#if defined(HAVE_NCURSESW_CURSES_H) || defined(HAVE_NCURSESW_H)
 				       "ncursesw",
 #else
 				       "ncurses",


krobelus (approved):

src/tig.c
```diff
@@ -557,7 +557,7 @@ parse_options(int argc, const char *argv[], bool pager_mode)
 				printf("tig version %s\n", TIG_VERSION);
 #ifdef NCURSES_VERSION
 				printf("%s version %s.%d\n",
-#ifdef NCURSES_WIDECHAR
+#ifdef HAVE_NCURSESW_CURSES_H
```
koutcher, line 560: Thanks for reporting and fixing the problem. To be consistent with `ax_with_curses.m4` we probably have to make it `#if defined(HAVE_NCURSESW_CURSES_H) || defined(HAVE_NCURSESW_H)`, don't you think ?

mqudsi: I believe you're correct and I've force-pushed an update that uses both. I'm not sure how I missed that.

PR #1243, opened 2022-10-27
Can edit in Pager

crisisdice:
When working with pull requests, before merging to `master` it is nice to get the total diff of all of the commits of the PR, as in the `Files changed` tab on github. This is possible with `tig`, i.e. `git diff master | tig` or `:!git diff master`, but currently, this view does not support editing.

This PR makes editing in the pager view possible.

### DIFF
--- a/src/pager.c
+++ b/src/pager.c
@@ -155,6 +155,9 @@ pager_request(struct view *view, enum request request, struct line *line)
 	enum open_flags flags = view_is_displayed(view) ? OPEN_SPLIT : OPEN_DEFAULT;
 	int split = 0;
 
+	if (request == REQ_EDIT)
+		return diff_common_edit(view, request, line);
+
 	if (request != REQ_ENTER)
 		return request;
 


src/pager.c
```diff
@@ -155,6 +155,9 @@ pager_request(struct view *view, enum request request, struct line *line)
 	enum open_flags flags = view_is_displayed(view) ? OPEN_SPLIT : OPEN_DEFAULT;
 	int split = 0;
 
+	if (request == REQ_EDIT)
+		return diff_common_edit(view, request, line);
```
krobelus, line 159: Seems like a good change, thanks. This has bothered me occasionally in the past, although I hardly use the pager view (perhaps because of that?).

One thing that's weird in `git log -p | tig` is that if my cursor is inside a commit message,
`:edit` will work - searches upwards for the last file in above commit.
I think we should stop the search at a commit boundary, e.g. the `commit:` line. I guess this is not a blocker.

Can we also make it work for the log view (e.g. `tig log -p`)?

One more thing: could you put the PR description in the commit message instead? The commit message will make it into the changelog (via `git shortlog`)

koutcher:
Thanks. It will only work when paging diff content but the worst that can happen is a `Nothing to edit` error message.

PR #1249, opened 2022-12-27
Highlight matching regex pattern in pager, via option color-pager-regex

psprint:
A third attempt after #1223 to commit flexible pager view output highlighter.

Any line matching the regex given by option `color_pager_regex` will be drawn in a new color ↔ `color-match`. The option is empty by default, as requested. The color is magenta by default.

Example:
2022-12-27-150236_1888x926_scrot

Typically useful to make compiler warnings easy visible.

### DIFF
--- a/include/tig/line.h
+++ b/include/tig/line.h
@@ -22,6 +22,7 @@ struct ref;
  */
 
 #define LINE_INFO(_) \
+        _(COLOR_MATCH,     ""), \
 	_(DIFF_HEADER,		"diff --"), \
 	_(DIFF_DEL_FILE,	"--- "), \
 	_(DIFF_ADD_FILE,	"+++ "), \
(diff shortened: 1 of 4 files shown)


psprint:
Hi,
Could be the PR be merged? It's an useful feature, I'm just used it to highlight a preamble of introducionary lines in a Tig run as the item chooser (like fzf), which is nice as the UI is the same as Tig.

koutcher:
Once again you reinvent the wheel, Tig already has a coloring engine that can be extended to support regular expressions... Look for the updated syntax in the commit message or in tigrc.5.adoc.

psprint:
The new support is great, I was able to colorize output of my new `tigsuite`s  feature plugin:

It allows to create project and subproject file lists that can be opened from `tig`:
2023-01-07-163841_1889x1045_scrot

However, it seems that the color always spans whole line. Would it be hard to colorize different parts of lines in different color?

PR #1255, opened 2023-01-17
fix some implicit declarations

nekopsykose:
fixes implicit-function-declaration of strcasecmp (strings.h) and implicit define of struct timezone (bsd/gnu source)

### DIFF
--- a/include/tig/tig.h
+++ b/include/tig/tig.h
@@ -44,6 +44,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <strings.h>
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -11,6 +11,8 @@ m4_pattern_forbid([^AX_])
 
 AC_PROG_CC
 
+AC_DEFINE(_BSD_SOURCE, 1, [Required for struct timezone])
```
koutcher, line 14: Thanks. As configure is not the only way to compile Tig, I'd rather have the define done in include/tig/tig.h:37.

nekopsykose: updated, right below the config.h include

koutcher: Thanks. Looks like we also need to add an extra `#define _DEFAULT_SOURCE` for glibc >= 2.19.

nekopsykose: ah, right

PR #1268, opened 2023-02-25
Fix "dangling pointer" warning

proski:
Move file_status to the function scope so it doesn't go out of scope while file_in_head is pointing to the data it contains.


proski:
This is one of the two warnings reported by gcc 12.2.1 with the `-Wall` flags. And it's an actual problem. When a variable goes out of scope its memory can be reused so its contents should not be used beyond than point. That is a well known problem in C++ where objects going out of scope call their destructors. But I can imagine it can be problematic in C as well if the compiler allocates stack exactly as written in the code.

PR #1269, opened 2023-02-25
Don't use array size in function arguments

proski:
The declaration of argv_to_string() mismatched its definition. The definition had the array size, the declaration had not. Remove array sizes from all functions declared in argv.h.

Array sizes in function arguments don't ensure the code correctness but can give a wrong impression that the compiler would check the array size.

### DIFF
--- a/include/tig/argv.h
+++ b/include/tig/argv.h
@@ -25,9 +25,9 @@
 
 bool argv_to_string(const char *argv[], char *buf, size_t buflen, const char *sep);
 char *argv_to_string_alloc(const char *argv[], const char *sep);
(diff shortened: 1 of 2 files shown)


proski:
I've checked that the code does correct checks for the argv size. The `argv_to_*` functions look for the NULL terminator. The `argv_from_*` functions fail if the resulting NULL terminated array would exceed `SIZEOF_ARG` in size. Both looks correct to me.

This is a very visible warning detected with `-Wall` both by gcc and clang. It's about time that we fix it so that `-Wall` can be enabled in all builds by default.

koutcher:
LGTM, thanks.

PR #1284, opened 2023-05-21
Allow to go to stage view without Enter

krobelus:
After selecting unstaged or staged changes in main or status view
and pressing "c" to go to stage view, we fail with this error

	No stage content, press s to open the status view and choose file

We can work around this by pressing Enter before "c".  This extra key
press is really not necessary because the intent is clear.  Let's make
view-stage (and view-diff) go to the stage view straight away.

Note: "d" already worked this way but only in the main view, I'm not
sure why.

The implementation needs to differentiate between "stat headers"
like "Changes to be committed" and status lines that actually contain
a file. For stat headers we show all files so we must be careful not
to include a file filter.
For untracked files, we show a blob so there is no natural way of
showing all untracked files. Keep the above behavior for the untracked
stat header.

### DIFF
--- a/include/tig/stage.h
+++ b/include/tig/stage.h
@@ -20,7 +20,7 @@ struct status;
 
 extern struct view stage_view;
 
-void open_stage_view(struct view *prev, struct status *status, enum line_type type, enum open_flags flags);
+void open_stage_view(struct view *prev, struct status *status, enum open_flags flags);
 
 #endif
 /* vim: set ts=8 sw=8 noexpandtab: */
--- a/include/tig/view.h
+++ b/include/tig/view.h
@@ -49,6 +49,7 @@ struct line {
 	unsigned int no_commit_refs:1;
 	unsigned int graph_indent:1;
 	unsigned int search_result:1;
+	unsigned int stat_header:1;
 
 	void *data;		/* User data */
 };
--- a/src/main.c
+++ b/src/main.c
@@ -92,6 +92,8 @@ main_add_commit(struct view *view, enum line_type type, struct commit *template,
 	line = add_line_alloc(view, &commit, type, titlelen, custom);
 	if (!line)
 		return NULL;
+	if (custom)
+		line->stat_header = 1;
 
 	*commit = *template;
 	strcpy(commit->title, title);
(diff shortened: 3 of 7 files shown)


koutcher:
Thanks @krobelus, but that's quite a lot of changes. In the spirit of what jonas did for "d" in the main view, couldn't we do more simply
```patch
diff --git a/src/main.c b/src/main.c
index 2401295b..05145955 100644
--- a/src/main.c
+++ b/src/main.c
@@ -568,6 +568,14 @@ main_request(struct view *view, enum request request, struct line *line)
                        open_diff_view(view, flags);
                break;

+       case REQ_VIEW_STAGE:
+               if (line->type == LINE_STAT_UNSTAGED
+                   || line->type == LINE_STAT_STAGED)
+                       open_stage_view(view, NULL, line->type, flags);
+               else
+                       return request;
+               break;
+
        case REQ_REFRESH:
                load_refs(true);
                refresh_view(view);
diff --git a/src/status.c b/src/status.c
index 0855dc34..7b476550 100644
--- a/src/status.c
+++ b/src/status.c
@@ -752,6 +752,15 @@ status_request(struct view *view, enum request request, struct line *line)
                        view->env->ref[0] = 0;
                return request;

+       case REQ_VIEW_STAGE:
+               if ((line->type == LINE_STAT_STAGED
+                    || line->type == LINE_STAT_UNSTAGED
+                    || line->type == LINE_STAT_UNTRACKED) && status && status->status)
+                       open_stage_view(view, status, line->type, view_is_displayed(view) ? OPEN_SPLIT : OPEN_DEFAULT);
+               else
+                       return request;
+               break;
+
        case REQ_ENTER:
                /* After returning the status view has been split to
                 * show the stage view. No further reloading is
```

krobelus:
Interesting. This doesn't work when running `:view-stage` on individual files but maybe there is a way to add it to this approach

koutcher:
Actually, in the status view, the result is better with just:
```patch
diff --git a/src/status.c b/src/status.c
index dd8c2afa..c9cfb684 100644
--- a/src/status.c
+++ b/src/status.c
@@ -756,6 +756,7 @@ status_request(struct view *view, enum request request, struct line *line)
                        view->env->ref[0] = 0;
                return request;

+       case REQ_VIEW_STAGE:
        case REQ_ENTER:
                /* After returning the status view has been split to
                 * show the stage view. No further reloading is
```

krobelus:
Thanks, that works as expected. I had forgotten to test it

Issue #1294, opened 2023-08-29
Chinese (utf-8) character display problem


echoonly:
Hi，
I have met a problem with the display of Chinese (utf-8) characters. The Chinese contained in the commit message or code will be displayed as a `~` symbol when viewed in tig:
![iShot_2023-08-29_17 04 01](https://github.com/jonas/tig/assets/11766632/85a6818f-eab9-4510-bdc6-914491deb017)

`git log` command output is normal:
![iShot_2023-08-29_17 17 53](https://github.com/jonas/tig/assets/11766632/cc31ade3-3604-48f8-bd35-f819b404c92c)

I checked the previous issue, which mentioned that it is necessary to link to the ncursesw to support UTF-8 character display. 
But, I have compiled tig with ncursesw using git source, `tig --version` output is:
![image](https://github.com/jonas/tig/assets/11766632/2f8f1e44-3e63-4a61-983b-a1a3db6228ce)

The compilation command:
```
make configure && ./configure --prefix=${DEST_PATH}/tig
make -j 4 && make install
```
The output related to ncursesw when the configure command is executed is:
![image](https://github.com/jonas/tig/assets/11766632/a3faec49-8e70-49ba-94e5-ce85e57e3b61)

Is there anything wrong with this process?  Thank you!

koutcher:
The process looks correct. It seems not all characters are affected as the third and last are ok, so the problem could be somewhere else. Centos/RHEL 7 I believe, based on the libraries versions. Do you have an example of accessible repository where you see the problem ?

echoonly:
Yes, my os is Centos 7
```
LSB Version:    :core-4.1-amd64:core-4.1-noarch
Distributor ID: CentOS
Description:    CentOS Linux release 7.6.1810 (Core)
Release:        7.6.1810
Codename:       Core
```

> Do you have an example of accessible repository where you see the problem ?

I don't quite understand what example you need.

PR #1299, opened 2023-10-24
Fix display non-ascii character.

nonakap:
Should cast the argument passed to is*() function in ctype.h to unsigned char.

See https://wiki.sei.cmu.edu/confluence/display/c/STR37-C.+Arguments+to+character-handling+functions+must+be+representable+as+an+unsigned+char.

Should fix https://github.com/jonas/tig/issues/1294.

before:
The commit log contains Japanese characters, but they are not displayed correctly.
tig-display-broken-non-ascii-character

after:
The commit log is displayed correctly.
tig-fix-display-non-ascii-character

### DIFF
--- a/src/string.c
+++ b/src/string.c
@@ -25,7 +25,7 @@ string_isnumber(const char *str)
 	int pos;
 
 	for (pos = 0; str[pos]; pos++) {
-		if (!isdigit(str[pos]))
+		if (!isdigit((unsigned char)str[pos]))
 			return false;
 	}
 
@@ -38,7 +38,7 @@ iscommit(const char *str)
 	int pos;
 
 	for (pos = 0; str[pos]; pos++) {
-		if (!isxdigit(str[pos]))
+		if (!isxdigit((unsigned char)str[pos]))
 			return false;
 	}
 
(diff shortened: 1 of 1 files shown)


koutcher:
Thank you @nonakap. Your proposed correction has the exact opposite effect on macOS, it works without and it doesn't work with, but it certainly goes in the right direction. Could you try the following ?
```patch
diff --git a/src/string.c b/src/string.c
index a90fc1bb..4cb0910b 100644
--- a/src/string.c
+++ b/src/string.c
@@ -103,7 +103,7 @@ string_expand(char *dst, size_t dstlen, const char *src, int srclen, int tabsize
                                expanded = dstlen - size - 1;
                        memcpy(dst + size, "        ", expanded);
                        size += expanded;
-               } else if (isspace((unsigned char)c) || iscntrl((unsigned char)c)) {
+               } else if (iswspace(c) || iswcntrl(c)) {
                        dst[size++] = ' ';
                } else {
                        dst[size++] = src[pos];
@@ -119,7 +119,7 @@ string_trim_end(char *name)
 {
        int namelen = strlen(name) - 1;

-       while (namelen > 0 && isspace((unsigned char)name[namelen]))
+       while (namelen > 0 && iswspace(name[namelen]))
                name[namelen--] = 0;

        return name;
@@ -128,7 +128,7 @@ string_trim_end(char *name)
 char *
 string_trim(char *name)
 {
-       while (isspace((unsigned char)*name))
+       while (iswspace(*name))
                name++;

        return string_trim_end(name);
```

PR #1303, opened 2023-11-04
Fix typos and update tools/aspell.dict

kianmeng:
Found via `make spell-check`


HaeyoonJo:
@kianmeng , good catch. IMHO, when we open PR, you need to commit in a single commit. See if this might be helpful in the following [Organizing PRs and Commits document](https://cockroachlabs.atlassian.net/wiki/spaces/CRDB/pages/1411744698/Organizing+PRs+and+Commits#:~:text=We%20recommend%20one%20commit%20per,review%2C%20to%20answer%20reviewer%20comments.)

PR #1310, opened 2023-12-27
Use HTTPS for GitHub clone URLs

jwilk:
GitHub has dropped support for the git:// protocol:
https://github.blog/changelog/2022-03-15-removed-unencrypted-git-protocol-and-certain-ssh-keys/


koutcher:
Thanks.

Issue #1315, opened 2024-02-11
blame view keybindings are hard to discover


ilyagr:
I wish they were documented in the in-program help better. There could be a "blame bindings" section. I realize that all the bindings are mentioned in the help already, but their descriptions make it very difficult to understand how they actually work in blame view. Blame is very useful occasionally, but I find it difficult to remember the bindings whenever I use it.

My understanding is that, in blame view:

- `Enter` opens the blamed commit for the current line in a separate pane
- `b` opens the current file at the blamed commit when already in blame view
- `<` goes back to the previous view, e.g. undoes `b` (as elsewhere in tig, but I only use it when blaming)
- `,` seems like it could be useful, but isn't in blame view

Thanks for making `tig`, it's very useful! I was thinking of making a PR for this, but I couldn't immediately figure out how the help is generated.

ilyagr:
I realized that even here, I misunderstood something. It seems that the `,` command is important in blame. I'm guessing it's meant to go to either (the last commit before the current one that modified the current file) or (the last commit before the blamed commit for the current line that modified the selected line).

One confusing thing about it is that it sometimes shows the message "The selected commit has not parents". I wish this message was clearer: is the selected commit the one on the current line or the one currently being viewed? It should also mention say something like "...has no ancestors that modify this (file/line)".

ilyagr:
BTW, I'd be happy to try implementing this myself, but it'd help if you gave me some pointers.

krobelus:
> , shows the last commit before the blamed commit for the current line that modified the selected line.

Right. I never really used, it seems nice. Sadly it can't know where a line moved to, so the initial position can be confusing.
I usually alternate between blame and diff view for recursive blaming.

ilyagr:
https://github.com/jonas/tig/issues/1315

> Sadly it can't know where a line moved to, so the initial position can be confusing.

That's a good point. If the "going to an ancestor of the currently displayed commit" operation existed, it'd be equivalent to `b`  followed by that operation, so it overall seems less important.

I think having `b`, `B` (optional, would do what `,` currently does), `,` (for going to a parent of the currently displayed commit, as seems to match its description in the help, and is nice to track the evolution of a file), and `<` would be a good interface, especially it's documented. The (new) blame section of the help could document `b` and `B`, and either also document `,` and `<` (repeating an earlier section, possibly with variations), or just mention "See also `,` and `<` commands which are particularly useful in blame view".

[...]

PR #1316, opened 2024-02-13
Blame: improve error messages for the `,` operation

ilyagr:
This addresses a portion of https://github.com/jonas/tig/issues/1315.

The message is now:

   Commit f4657f00 has no ancestors modifying the selected line

instead of  

   The selected commit has no parents

Apart from being a bit clearer, the user will be able to learn what `,`
is meant to do from it. In particular, they should notice that commit
`f4657f00` is the commit on the selected *line*, not the commit that is
currently displayed in general.

Trying to press `,` a few times is a natural thing to try to do after
reading the help while trying to figure out how blame works in tig.

----------

I would still appreciate pointers on how to create a help section. I'm also
curious whether you think that the additional command I describe in the "update" in https://github.com/jonas/tig/issues/1315#issuecomment-1942506665 is a good idea.

I'd also need a pointer if you'd like me to add the test. Changing the message didn't seem to break any tests.

### DIFF
--- a/src/blame.c
+++ b/src/blame.c
@@ -355,12 +355,12 @@ blame_go_forward(struct view *view, struct blame *blame, bool parent)
 	const char *filename = parent ? commit->parent_filename : commit->filename;
 
 	if (!*id && parent) {
-		report("The selected commit has no parents");
+		report("Commit %.8s has no ancestors modifying the selected line", commit->id);
 		return;
 	}
 
 	if (!strcmp(history_state->id, id) && !strcmp(history_state->filename, filename)) {
-		report("The selected commit is already displayed");
+		report("Commit %.8s is already being displayed", commit->id);
 		return;
 	}
 


src/blame.c
```diff
@@ -355,12 +355,12 @@ blame_go_forward(struct view *view, struct blame *blame, bool parent)
...
 
 	if (!*id && parent) {
-		report("The selected commit has no parents");
+		report("Commit %.8s has no ancestors modifying the selected line", commit->id);
 		return;
 	}
 
 	if (!strcmp(history_state->id, id) && !strcmp(history_state->filename, filename)) {
-		report("The selected commit is already displayed");
+		report("Commit %.8s is already beind displayed", commit->id);
```
krobelus, line 363: typo: being.
Also I wonder how to hit this case?

> I would still appreciate pointers on how to create a help section

The existing help text is generated for commands, I guess we don't want to add commands.
Probably we can add some lines that are not associated with a command. Not completely trivial but doable.

>  I'm also
curious whether you think that the additional command I describe in the "update" in https://github.com/jonas/tig/issues/1315#issuecomment-1942506665 is a good idea.

Having a command that shows the previous version of a file doesn't seem useful very often, unlike recursive the commands that use the commit that added the line at cursor.

ilyagr: Thanks for looking this over!

> typo: being.

Fixed, thanks for catching it!

> Also I wonder how to hit this case?

No idea, I was wondering that myself.

> The existing help text is generated for commands, I guess we don't want to add commands.

I imagine we could add commands with the same keybindings only for the blame view. I'm not sure whether they could have the same names (for people who reconfigured things), though.

> Probably we can add some lines that are not associated with a command. Not completely trivial but doable.

This would be great! Also, is the man page generated from the commands as well?

> Having a command that shows the previous version of a file doesn't seem useful very often, unlike recursive the commands that use the commit that added the line at cursor.

I think it's useful for a few reasons:

- To browse the history of one file. I am not sure whether Tig has that functionality already, the closest I found is the "blob view", but it doesn't support the `,` command.
- As I'm about to post in #1315, together with `b` such a command can replace `,`'s current behavior, but not vice-versa.
- For consistency with what the command does in other screens (I assume, I haven't double-checked)
- For consistency with how the command is described in help (unless we can think of a better description)

ilyagr: > Having a command that shows the previous version of a file doesn't seem useful very often, unlike recursive the commands that use the commit that added the line at cursor.

In #1315, you pointed out that you use the diff view for this. Perhaps the reason I want this command is that I don't know how to do that effectively, especially if I want to focus on one file.

krobelus: I usually don't really care about the history of a full file[*] but about a single line (and the related changes).

> I usually alternate between blame and diff view for recursive blaming.
> 
> I'm realizing I'm not sure how to do this,

Often a line has been modified by several (uninteresting) commits.
So what I (used to) do is
1. press "b"
2. press Enter to show the diff that introduced the line
3. use "/foo" and/or "n" to select the old version of the line (the one prefixed with a "-")
4. go back to step 1 (until I have found the interesting commit)

There are many alternatives that can help like the view-parent, using an
ignore-revs-file using `git blame -w` to ignore whitespace changes, using
`tig -Gfoo`.. but I find the above steps to be the most general one and
least likely to break (assuming your Git history has some level of continuity).

Probably there is room for using `,`

I've recently switched to the dark side and [implemented this workflow in my editor](https://github.com/mawww/kakoune/commit/86d940c225d4d1e316efebf6d915feaa1aa49f63), with one neat speedup:
My step 2 (pressing Enter to show the blamed commit) will not move the viewport to the top of the commit.
Instead it jumps directly to the line that is being blamed.
Since in most cases, the old version of the line is right next to the new version (which is the blamed one), this usually makes step 3 trivial.

I'd like to have this in Tig as well (I still use it for heavy-duty Git operations).
Perhaps a B shortcut (`view-blame-jump`) that's like `view-blame` but moves to the blamed line. Or change `view-blame`

---

[*] If I care about the history of a file, I don't think I'd want to look
at it by eyeballing blob views. Rather I'd look at the diffs from `tig my-files.c`.

> For consistency with what the command does in other screens (I assume, I haven't double-checked)

Normally, `view-parent` goes to the parent commit so that's not the same as "youngest ancestor that touched this file" anyway.
And that's hardly more intuitive than "youngest ancestor that touched this line" even though I agree that one is confusing too

ilyagr: > I'd like to have this in Tig as well (I still use it for heavy-duty Git operations).

That would be very nice!

I appreciate you adding this to Kakoune, too; I'll try it out once the new release comes out. I do usually use VS Code + Dance plugin, though, so I'd have to open Kakoune specially just for this in most cases.

Also, I'm currently wishing not just for a tool I can use, but a tool I can recommend to others. Kakoune is not it, unfortunately, at least not for this purpose.

> Normally, view-parent goes to the parent commit so that's not the same as "youngest ancestor that touched this file" anyway.

I'm not sure how valid your argument is; it depends on whether in the rest of `tig`, `view-parent` goes to the parent commit of the current *view* or of the *selected line*. I'm hoping it's the former, in which case I think my suggested behavior is consistent.

But in any case, the strongest argument for changing its behavior, IMO, is that you can easily get the line behavior from the "this file" behavior by combining it with `b`, it just becomes `b<` instead of `<`. You can't get the behavior I want at all, I think, in the current `tig`. I want it one way or another, and the other (current) behavior could either be `b<` or it could be a separate key.

koutcher: Thanks for your proposal. I don't think the problem really lies in the error message but rather in the lack of information in the title window (to which the error message is supposed to refer). I tried to improve this in https://github.com/koutcher/tig/commit/b1cda8cad6b71e3b8c294f2bd9c9a10f9c88299a. 

Both keys are properly described in the man page, but I agree it is not always easy to understand where the annotating starts after using `b` or `,`. To make it clearer, I also added a notification in the status window when the starting point changes. HTH

ilyagr: Thank you for working on this! I posted some thoughts at https://github.com/koutcher/tig/commit/b1cda8cad6b71e3b8c294f2bd9c9a10f9c88299a. I certainly think your commit is an improvement, but I think it'd be nice to clarify its messages further, if possible.

PR #1318, opened 2024-03-04
refdb: ignore prefetch refs

arbrauns:
These are created by `git maintentance` and are special-cased in git tools to not be shown, since they are mostly noise.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -12,8 +12,8 @@ Improvements:
  - Use HTTPS for GitHub clone URLs. (#1310)
  - Move default log view options to tigrc.
  - Allow to go to stage view without Enter. (#1284)
- - Add new "prefetch" reference type for refs created by `git maintenance`.
-   (#1318)
+ - Add new "prefetch" reference type for refs created by `git maintenance`
+   (hidden in default config). (#1318)
 
 Bug fixes:
 
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -12,6 +12,8 @@ Improvements:
  - Use HTTPS for GitHub clone URLs. (#1310)
  - Move default log view options to tigrc.
  - Allow to go to stage view without Enter. (#1284)
+ - Add new "prefetch" reference type for refs created by `git maintenance`.
+   (#1318)
 
 Bug fixes:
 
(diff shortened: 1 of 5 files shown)


Gerschtli:
@arbrauns Do you know a workaround for that issue other than using your forked version?

arbrauns:
No, I'm not aware of any existing filtering that could be used for this.

koutcher:
With Tig >= 2.5.5, add a line `set reference-format = hide:other` to your `.tigrc`.

NEWS.adoc
```diff
@@ -12,6 +12,8 @@ Improvements:
  - Use HTTPS for GitHub clone URLs. (#1310)
  - Move default log view options to tigrc.
  - Allow to go to stage view without Enter. (#1284)
+ - Add new "prefetch" reference type for refs created by `git maintentance`
```
koutcher, line 15: typo

doc/tigrc.5.adoc
```diff
@@ -179,9 +179,12 @@ The following variables can be set:
...
 	names. Wrap the name of the reference type with the characters you would
-	like to use for formatting, e.g. `[tag]` and `<remote>`. If no format is
-	specified for `local-tag`, the format for `tag` is used. Similarly, if no
-	format is specified for `tracked-remote` the `remote` format is used.
+	like to use for formatting, e.g. `[tag]` and `<remote>`.
+	- If no format is specified for `local-tag`, the format for `tag` is used.
+	- If no format is specified for `tracked-remote`, the format for `remote`
+	  is used.
+	- If no format is specified for any other reference type, the format for
+	  `branch` is used.
```
koutcher, line 187: This breaks the layout of the generated man page. Separating list in asciidoc while staying in the same block doesn't look straightforward.

koutcher (changes requested):
Thanks, the code part is OK but the documentation can't be merged as is.

doc/tigrc.5.adoc
arbrauns: I can't get it to work with the legacy https://pypi.org/project/asciidoc/ either, but asciidoctor can handle it - switching the man pages to that is probably a more complicated endeavour though. I've reverted it back to the running text style.

PR #1326, opened 2024-04-07
Fix incorrect filename coloring in `tig status`

xaizek:
Due to the way commit a37ce5cfe1f256b81092dda365b1d66450e0929e updated src/draw.c:draw_filename() the whole line in `tig status` got colored according to `stat-*` which supposed to color only status markers.

***

I don't see other such adhoc conditionals in the file, so there is probably a better way of doing it, but I thought this demonstrates the issue and the approximate way of addressing it.

### DIFF
--- a/NEWS.adoc
+++ b/NEWS.adoc
@@ -1,6 +1,14 @@
 Release notes
 =============
 
+tig-2.5.10
+----------
+
+Bug fixes:
+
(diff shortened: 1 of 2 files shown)


koutcher:
Good catch, thanks for the proposition. I think I'd rather revert the changes in draw_filename() as they were made only to match changes draw_filesize() where they were needed to skip the display of size for submodules.

PR #1337, opened 2024-05-22
diff parsing, regardless of diff-prefix choices

### LINKED ISSUE #1336: bug: fail to find filename if not showing prefixes
1. `git config diff.noprefix true`
2. open `tig blame`
3. select a diff section and press <kbd>b</kbd>

what i expect:
To be magically taken to the previous review on the diff.

what happens:
> Failed to read the file name

(very minor bug. mostly to document to future me when i see that error again)

gcb:
handle all diff view options (noprefix, mnemonicPrefix, srcPrefix, dstPrefix, line-prefix)

(moved quick fix of #1336 to #1338)

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -657,7 +657,7 @@ diff_trace_origin(struct view *view, struct line *line)
 		if (!prefixcmp(data, "--- ")) {
 			const char *data_n1 = box_text(diff+1);
 			const char *data_n2 = box_text(diff+2);
-			if(!prefixcmp(data_n1, "+++ ") && prefixcmp(data_n2, "@@ ")){
+			if(!prefixcmp(data_n1, "+++ ") && !prefixcmp(data_n2, "@@ ")){
 				file = data + STRING_SIZE("--- ");
 				break;
 			}
--- a/src/diff.c
+++ b/src/diff.c
@@ -654,13 +654,10 @@ diff_trace_origin(struct view *view, struct line *line)
 
 	for (; diff < line && !file; diff++) {
 		const char *data = box_text(diff);
+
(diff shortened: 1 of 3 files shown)


src/diff.c
```diff
@@ -654,10 +654,13 @@ diff_trace_origin(struct view *view, struct line *line)
 
 	for (; diff < line && !file; diff++) {
 		const char *data = box_text(diff);
-
-		if (!prefixcmp(data, "--- a/")) {
-			file = data + STRING_SIZE("--- a/");
-			break;
+		if (!prefixcmp(data, "--- ")) {
+			const char *data_n1 = box_text(diff+1);
+			const char *data_n2 = box_text(diff+2);
```
gcb, line 659: probably needs a bounds check here

koutcher:
7h4nk5 for reporting and trying to fix the issue. Test case is failing because your fix breaks the original prefixed use case.

gcb:
Thanks for pointing out the test issue! will try to polish this when i get some down time next month, if nobody beat me to it.

src/diff.c
```diff
@@ -654,10 +654,13 @@ diff_trace_origin(struct view *view, struct line *line)
...
+			if(!prefixcmp(data_n1, "+++ ") && !prefixcmp(data_n2, "@@ ")){
+				file = data + STRING_SIZE("--- ");
```
gcb: all these hardcoded strings can probably be from https://github.com/jonas/tig/blob/master/include/tig/line.h#L29

gcb:
also needs to account for `mnemonicPrefix`, which can change prefix letters.

PR #1338, opened 2024-05-30
fix diff blame navigation

### LINKED ISSUE #1336: bug: fail to find filename if not showing prefixes

gcb:
this does **not** "support" git options such as `noprefix`, `mnemonicPrefix`, etc. As in, they will not have the desired effect on diff view, like #1337 was trying to implement.

It just overrides them with the `--default-prefix` command line when getting the diffs, so blame navigation (e.g. pressing "b" in a diff) works.

fix #1336

### DIFF
--- a/src/diff.c
+++ b/src/diff.c
@@ -27,7 +27,7 @@ diff_open(struct view *view, enum open_flags flags)
 {
 	const char *diff_argv[] = {
 		"git", "show", encoding_arg, "--pretty=fuller", "--root",
-			"--patch-with-stat", use_mailmap_arg(),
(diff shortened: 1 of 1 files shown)


gcb:
About the failing tests... can either remove the existing places where diff prefixes are handled, or try to centralize them. e.g. https://github.com/jonas/tig/blob/tig-2.5.10/src/diff.c#L745 (there are others, searching for "prefix" on the log)

Issue #1347, opened 2024-07-17
macOS search ('/') is not <Ctrl-C> cancellable


landonb:
On macOS, when I start a search (using '/'), I cannot cancel out of it.

- Pressing `<Ctrl-C>` does nothing.

It seems like the `readline` call ignores `SIGINT`.

The only way I can return to the main view is to press `<Enter>` (and to run the search).

- Interestingly, I can `kill -s 2 $(pgrep tig)` from another terminal window, and that will cancel the search.

I do not experience this issue on Linux Mint MATE.

### OS and `tig` versions

The problem affects both Brew `tig` and my own `tig` on macOS Sonoma 14.5:

```
  @macOS $ /opt/homebrew/Cellar/tig/2.5.10/bin/tig --version
  tig version 2.5.10
  ncursesw version 6.5.20240427
  readline version 8.2
  PCRE2 version 10.44 2024-06-07
```

Here's my own build, using the [latest commit](https://github.com/jonas/tig/commit/37ccb67e95ed7e9fc9e49872f9b65f3ca692f614) (and the latest `readline` and `ncurses` from Brew):

```
  @macOS $ tig --version
  tig version 2.5.10-4-g37ccb67e
  ncursesw version 6.5.20240427
  readline version 8.2
  PCRE2 version 10.44 2024-06-07
```

- I can attach my configure and build commands if you'd like to see them.

I do not see the problem on Linux Mint MATE ~21.4~ 21.3, also running the latest commit, albeit not the same `readline` (though it's the latest from `apt`):

```
  @linuxmint $ tig --version
  tig version 2.5.10
  ncursesw version 6.3.20211021
  readline version 8.1
  PCRE2 version 10.39 2021-10-29
```

### (Limited) diagnosis

I am somewhat surprised that I didn't find an existing issue regarding this, but I looked.

- I also don't think there's anything unique about my host. I can send `<Ctrl-C>` to the terminal itself and to other apps, and `<Ctrl-C>` works from `tig` main view.

I also haven't debugged C code in eons, and I'm not *that* familiar with `readline` sources, but I tracked it down to the [`prompt.c` `realine` call](https://github.com/jonas/tig/blob/master/src/prompt.c#L503)

```
  line = readline(prompt);
```

- Or at least that's the call that blocks until `<Enter>`, and the `sigint_absorb_handler` registered above it never gets called.

[...]

Issue #893, opened 2019-01-28
Piping to tig segfaults on exit


verdie-g:
```
$ tig --version 
tig version 2.4.1
ncursesw version 6.1.20180127
readline version 8.0
```

Steps to reproduce:
- echo | tig
- :q (to exit)

valgrind output:
```
Invalid read of size 4
   at 0x4893C81: _rl_update_final (in /usr/lib/libreadline.so.8.0)
   by 0x489A8F6: rl_newline (in /usr/lib/libreadline.so.8.0)
   by 0x487CA2D: _rl_dispatch_subseq (in /usr/lib/libreadline.so.8.0)
   by 0x487CFFF: readline_internal_char (in /usr/lib/libreadline.so.8.0)
   by 0x487D825: readline (in /usr/lib/libreadline.so.8.0)
   by 0x124B10: ??? (in /usr/bin/tig)
   by 0x125CEB: ??? (in /usr/bin/tig)
   by 0x1135F6: ??? (in /usr/bin/tig)
   by 0x494B222: (below main) (in /usr/lib/libc-2.28.so)
 Address 0x0 is not stack'd, malloc'd or (recently) free'd


Process terminating with default action of signal 11 (SIGSEGV): dumping core
 Access not within mapped region at address 0x0
   at 0x4893C81: _rl_update_final (in /usr/lib/libreadline.so.8.0)
   by 0x489A8F6: rl_newline (in /usr/lib/libreadline.so.8.0)
   by 0x487CA2D: _rl_dispatch_subseq (in /usr/lib/libreadline.so.8.0)
   by 0x487CFFF: readline_internal_char (in /usr/lib/libreadline.so.8.0)
   by 0x487D825: readline (in /usr/lib/libreadline.so.8.0)
   by 0x124B10: ??? (in /usr/bin/tig)
   by 0x125CEB: ??? (in /usr/bin/tig)
   by 0x1135F6: ??? (in /usr/bin/tig)
   by 0x494B222: (below main) (in /usr/lib/libc-2.28.so)
```

koutcher:
Same here after Homebrew updated readline to 8.0. The tig test suite also fails with the same error. Switched back to readline 7.0.5 for now.

`brew switch readline 7.0.5`
`brew reinstall tig --build-from-source`

PR #1348, opened 2024-07-17
Fix readline ignores Ctrl-C on macOS (#1347)

landonb:
Proposed PR for #1347

- On macOS, you cannot cancel out of search ('/') using Ctrl-C (SIGINT), but rather must press <Enter> (and perform the search) to return to the main view.

- This commit partially reverts ea43f8be, “Fix segfaults with readline 8.0” (and associated Issue, “Piping to tig segfaults on exit” #893), but just for macOS builds.


koutcher:
Dupe of #1342

PR #1365, opened 2025-02-16
Added GitHub CI/CD workflow to generate and release man pages

alberti42:
Thank you for the amazing utility.

I wanted to be able to install `tig` directly from your GitHub repo, but I wanted to install on all my linux machines asciidoc to generate the man pages. Thus, I added a GitHub CI/CD workflow that generates the man pages from the asciidoc and then release them automatically. This way, I can install `tig` directly using `zinit`. 

You do not need to understand the code below if you never worked with zinit. It is just to give you an idea how you can install `tig` directly. The instruction `dl` downloads the man pages from the GitHub release page.

```zsh
zinit ice \
  binary \
  depth=1 \
  wait \
  lucid \
  dl"https://github.com/alberti42/fork-tig/releases/download/latest-man-pages/tig.1 -> $ZINIT[MAN_DIR]/man1/tig.1;
  https://github.com/alberti42/fork-tig/releases/download/latest-man-pages/tigmanual.7 -> $ZINIT[MAN_DIR]/man7/tigmanual.7;
  https://github.com/alberti42/fork-tig/releases/download/latest-man-pages/tigrc.5 -> $ZINIT[MAN_DIR]/man5/tigrc.5;" \
  make \
  id-as'jonas/tig' \
  lbin'src/tig -> tig' \
  nocompile
zinit light @alberti42/fork-tig
```

I believe other people can find other useful applications of the automatically released man pages.

### DIFF
--- a/.github/workflows/generate-man-pages.yml
+++ b/.github/workflows/generate-man-pages.yml
@@ -1,0 +1,41 @@
+name: Generate Man Pages
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build-man-pages:
+    runs-on: ubuntu-latest
+    container:
+      image: asciidoctor/docker-asciidoctor
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Generate Man Pages
+        run: |
+          mkdir -p man_pages
+          asciidoctor -b manpage -o man_pages/tig.1 doc/tig.1.adoc
+          asciidoctor -b manpage -o man_pages/tigmanual.7 doc/tigmanual.7.adoc
+          asciidoctor -b manpage -o man_pages/tigrc.5 doc/tigrc.5.adoc
+          ls -l man_pages/
+
+      # - name: Upload Man Pages as Artifact
+      #   uses: actions/upload-artifact@v4
+      #   with:
+      #     name: man-pages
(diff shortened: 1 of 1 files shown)


alberti42:
Did you have any time to look into it? It should be a noninvasive, low-risk operation. It only adds new assets to your releases, including the man pages.

Let me know if you need changes/fixes.

koutcher:
Generated man pages are already accessible from the repo in the `release` branch (e.g. https://raw.githubusercontent.com/jonas/tig/refs/heads/release/doc/tig.1).

alberti42:
Thanks for your reply.

If you don't think it is valuable to have an automated CI/CD for the man pages to produce release artifacts, feel free to close the PR.

The pro: it is probably a no-risk decision to publish the man pages as an artifact together with the various releases. It happens automatically, and you would not have to think about it. This would ensure people download and use the correct man page for the particular released version they are using. In this sense, the published man pages are not to be seen as a duplicate. Artifacts in the release section are also not adding up to the Git commit history. So, I would only see positive reasons for publishing them separately.

Also, if you check my installation line using zinit, it was somehow involved, and I had to use a special extension of zinit to download individual documents from the repo (using the keyword `dl`). Arguably not so complicated, but I thought we could make the installation process even smoother if the man pages were available in the release.

But if you have other reasons or concerns, let me know.

Cheers,
Andrea

PR #1370, opened 2025-03-21
Use correct line from parent on recursive blame on blob

### LINKED ISSUE #1369: blame: The `b` key in blame view does not go to the correct line since ca0809dbd
To reproduce, run `tig blame +28 tig-2.5.12 -- src/diff.c` in the `tig` repo with `tig 2.5.12`, and then press `b`. This takes me to line 28 of commit a1894.

This is incorrect. The correct behavior would take me to the corresponding line 54 of commit a1894.

That is because [line 28 of `tig-2.5.12`](https://github.com/jonas/tig/blob/b5f6193d968faedb85c193da126d29b59b26cd67/src/diff.c#L28) is ` const char *diff_argv[] = {` and the blame commit is a1894.The corresponding line in commit a1894 is [line 54](https://github.com/jonas/tig/blame/a1894d17695db1f1b15a296b2423909c3d9e7430/src/diff.c#L54) (and not line 28).

I bisected this; the behavior was correct until commit ca0809dbd which broke it (while fixing https://github.com/jonas/tig/issues/973).

Cc @krobelus if you are interested.

With some pointers, I might be able to help fixing it, but it worries me that fixing this bug might break another behavior.

---------

I don't think it's relevant (**Update:** the bug is unchanged by setting `GIT_CONFIG=/dev/null`, so it's **not** relevant), but here is the tig-related portion of my gitconfig:

```
[tig]
mouse=true
refresh-mode=periodic
refresh-interval=3
main-view-date-display=relative-compact
main-view-id-display=true
blame-view=author:abbreviated date:relative-compact id:yes,color line-number:yes,interval=1 text
# Tig always shows "author dates" instead of "commit dates",
# making reflog dates are very confusing
reflog-view-date-display=relative
# Don't show refs made by `jj`, `git-branchless`, etc.
reference-format=hide:other

[tig "color"]
# cursor = default default underline
cursor = default default reverse dim
# cursor = black white
```

krobelus:
Running

	tig blame +28 tig-2.5.12 -- src/diff.c

and pressing "b" should keep the line containing "const char *diff_argv[]
= }" selected. It actually moves the selection somewhere else.

This regressed in ca0809db (Fix cursor position after "Move to parent"
in blame view, 2019-11-29).

Commit ca0809db was only about "," (":back" or ":parent").  When we press
"b" (":view-blame"), we want to use the old behavior (originally added in
ba7c7d30 (Use file and line number information when loading blame for commit,
2009-02-07)). Do that.

(Note that the behavior fixed by ca0809db regressed in 22807341 (Enable
textconv in the blame view and fix blame -L (#1190), 2022-06-03).  I have
not yet figured out what's going on, but this patch shouldn't make that worse.)

Fixes #1369

### DIFF
--- a/src/blame.c
+++ b/src/blame.c
@@ -373,9 +373,12 @@ blame_go_forward(struct view *view, struct blame *blame, bool parent)
 
 	string_ncopy(view->env->ref, id, sizeof(commit->id));
 	string_ncopy(view->env->file, filename, strlen(filename));
-	if (parent)
+	if (parent) {
 		setup_blame_parent_line(view, blame);
-	view->env->goto_lineno = view->pos.lineno;
+		view->env->goto_lineno = view->pos.lineno;
+	} else {
+		view->env->goto_lineno = blame->lineno - 1;
+	}
 	reload_view(view);
 }
 
--- a/test/blame/blob-blame-test
+++ b/test/blame/blob-blame-test
@@ -1,0 +1,55 @@
+#!/bin/sh
+
+. libtest.sh
+. libgit.sh
+
+export COLUMNS=200
+
+in_work_dir create_repo_from_tgz "$base_dir/files/scala-js-benchmarks.tgz"
+
+tigrc <<EOF
+set line-graphics = ascii
+EOF
+
+steps '
+	# Go to the "clean in deltablue" line.
+	:36
+	# Find the commit that added that line.
+	:view-blame
+	# Assert that we still select that line.
+	:save-display recursive-blame.screen
+'
+
+test_tig blame project/Build.scala
+
+assert_equals 'recursive-blame.screen' <<EOF
+90286e0 Jonas Fonseca 2013-10-14 14:56 -0400  22|           "-feature",
+90286e0 Jonas Fonseca 2013-10-14 14:56 -0400  23|           "-encoding", "utf8"
+90286e0 Jonas Fonseca 2013-10-14 14:56 -0400  24|       )
+90286e0 Jonas Fonseca 2013-10-14 14:56 -0400  25|   )
+90286e0 Jonas Fonseca 2013-10-14 14:56 -0400  26|
+90286e0 Jonas Fonseca 2013-10-14 14:56 -0400  27|   lazy val benchmarkSettings = defaultSettings ++ Seq(
+90286e0 Jonas Fonseca 2013-10-14 14:56 -0400  28|       unmanagedSources in (Compile, packageJS) +=
+90286e0 Jonas Fonseca 2013-10-14 14:56 -0400  29|           baseDirectory.value / "exports.js"
+90286e0 Jonas Fonseca 2013-10-14 14:56 -0400  30|   )
+90286e0 Jonas Fonseca 2013-10-14 14:56 -0400  31|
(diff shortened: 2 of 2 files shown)


ilyagr:
I should move this to a separate bug report, but here are a couple of debugging nuggets about the change of the behavior of `,` in 2280734.

The reproduction of the bug is exactly the same as #1369, except you press `,` instead of `b`.

- The line https://github.com/jonas/tig/blob/2280734104fd362a1188b0678109e01cdb9322cf/src/blame.c#L337 never gets executed. It does get executed in the parent commit.
- Even though the command in https://github.com/jonas/tig/blob/2280734104fd362a1188b0678109e01cdb9322cf/src/blame.c#L311 changed slightly in that commit, its output seems to remain exactly the same.

    In other words, removing `--no-textconv` in `git diff --encoding=UTF-8 --no-textconv --no-ext-diff --no-color -U0 ae8829311b62c1432e2d0b5118804a5b9c629c4a:src/diff.c a1894d17695db1f1b15a296b2423909c3d9e7430:src/diff.c --` makes no difference.

ilyagr:
After a bit more debugging, I haven't gotten to the bottom of why, but I think the fix might be:

```diff
diff --git a/src/blame.c b/src/blame.c
index 56278952ee..dc2888f815 100644
--- a/src/blame.c
+++ b/src/blame.c
@@ -332,7 +332,7 @@
 				blamed_lineno = atoi(pos + 1);
 
 		} else if (*line == '+' && parent_lineno != -1) {
-			if (blame->lineno == blamed_lineno - 1 &&
+			if (blame->lineno == blamed_lineno &&
 			    !strcmp(blame->text, line + 1)) {
 				view->pos.lineno = parent_lineno ? parent_lineno - 1 : 0;
 				break;
```

The main question is whether it's better to do this or to shift all the `blame->lineno`s by 1.

test/blame/b-key-navigation-test
```diff
@@ -0,0 +1,58 @@
...
+set line-graphics = ascii
+EOF
+
+steps '
+	# Press b to go to blame for that line/commit
+	:view-blame
+	:save-display recursive-blame.screen
+'
+
+test_tig blame +27 project/Build.scala
```
ilyagr, line 24: These aren't really problems, but I copied this test to make a test for the `,` key bug in #1372, and we could coordinate the two tests.

- If you use line 36 here, it works as an example for both `,` and `b` keys
- You could rename the file to something like `navigation-b-key-test`, so that the two tests get placed next to each other. (This seems slightly simpler than combining them into one test).

test/blame/b-key-navigation-test
```diff
@@ -0,0 +1,58 @@
...
+
+# This shows that after pressing 'b' we should be positioned the line starting
```
ilyagr, line 26: Also not really a problem, but this confused me at first. It'd help if you at least added the expected line number and the fact that it's shown in the status bar.

Also, there's a typo, "on" is missing after "positioned".

Here's the comment I've come up with for the corresponding spot in #1372:

> This shows that after pressing ',' we should be positioned on line 40 (starting with "clean in common"), as shown in the status bar. We should *not* stay on line 36.

src/blame.c
ilyagr, line 1: Very minor: the commit (and PR) title could say "...from ancestor" instead of "...from parent" to be clearer and less confusable with #1372.

PR #1371, opened 2025-03-22
fix the weird line shift after staging file

liangkarl:
There is a bug with an incorrect line shift. After entering the stage window and staging any file in the main view, you will find a weird line shift, jumping two lines instead of one.

## Reproduce steps
1. `touch a b c d`
2. Open tig and enter "Untracked changes"
3. Press `u` to stage any file 
4. You could find the line shifting two instead of one.

### DIFF
--- a/src/status.c
+++ b/src/status.c
@@ -652,7 +652,7 @@ status_update(struct view *view)
 		return false;
 	}
 
-	if (line->type != LINE_STAT_STAGED && !no_files_staged)
+	if (!show_untracked_only && line->type != LINE_STAT_STAGED && !no_files_staged)
 		view->pos.lineno += 1;
 
 	return true;


src/status.c
```diff
@@ -652,9 +652,6 @@ status_update(struct view *view)
 		return false;
 	}
 
-	if (line->type != LINE_STAT_STAGED && !no_files_staged)
```
koutcher, line 655: Good catch but wrong solution, this code is needed in the plain status view. However we should not call it in the "Untracked only" status view:

```patch
diff --git a/src/status.c b/src/status.c
index 1e79cb82..82a52bec 100644
--- a/src/status.c
+++ b/src/status.c
@@ -652,7 +652,7 @@ status_update(struct view *view)
                return false;
        }

-       if (line->type != LINE_STAT_STAGED && !no_files_staged)
+       if (!show_untracked_only && line->type != LINE_STAT_STAGED && !no_files_staged)
                view->pos.lineno += 1;

        return true;
```

liangkarl: Got it. Thanks for the explanation! : )

liangkarl: So, do I need to close the PR, or will you close it after committing the changes?

koutcher: Leave it open, I'll use it the commit the changes.

PR #1373, opened 2025-03-22
manual: add a section about navigating blame/diff views

ilyagr:
This functionality, in particular the relevance and usefulness of the `,` and `b` commands in the blame or diff views, was previously hard to discover except by experimentation. I hope documenting it will let more people take advantage of it.

See also https://github.com/jonas/tig/issues/1315.

### DIFF
--- a/doc/manual.adoc
+++ b/doc/manual.adoc
@@ -248,6 +248,30 @@ be appended:
 [main] 77d9e40fbcea3238015aea403e06f61542df9a31 - commit 1 of 779 loading 5s 0%
 -------------------------------------------------------------------------------
 
+[[blame]]
+Blame and diff navigation
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Tig allows quick navigation between blame views and related diff views.
+
+In the diff view, navigating to an added line in a diff and pressing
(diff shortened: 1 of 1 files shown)


doc/manual.adoc
```diff
@@ -248,6 +248,30 @@ be appended:
...
+Tig allows quick navigation between blame views and related diff views.
+
+In the diff view, navigating to an added line in a diff and pressing
+'b' will open the blame view of the commit that was shown and select
+the corresponding line in that view. Navigating to a removed line in the
+diff and pressing 'b' takes you to the blame view of the parent commit.
+
+The blame view can also be opened directly with the 'tig blame' command.
+
+In the blame view, the blame commit for the current line is considered
```
eMPee584, line 264: This is somewhat hard to understand (huh, blame commit for the current line?), could you maybe try to reword it?
Also, when I move from line to line (in blame view), nothing changes in the status bar for me (also tested with `TIGRC_SYSTEM=` to skip local config..), maybe you can provide a screenshot of what you mean by that?
Otherwise, this looks good, and thanks for teaching us more TUI hacks ;  )

PR #1375, opened 2025-04-03
Open diff view from blame at the correct line

ilyagr:
This could be considered an RFC or even just a feature request. The feature goes with #1372 and #1370, but AFAICT it was never implemented before.

The implementation is a bit rough at this point; I had trouble following Tig's structure (I'm not an expert C programmer), would need help to make the code fit in better and maybe be more efficient.

However, it seems to work fine. Even if this is not merged, people can use it as a patch.

------------

When pressing <Enter> (AKA `:enter`) on a line in the blame view,
previously the diff view would open with the cursor either on the first
line or on whatever line was last selected in that diff view (if it was
previously opened). This commit makes the diff view open at the line
corresponding to the line that was selected in the blame view.

One can still press <d> (AKA `:view-diff`) to open the diff view at
either the first line or the previously selected line.

The implementation is currently proof-of-concept quality: it works,
is tested, but does not fit well with the rest of the code. I'd need
help to make the memory management more robust and to fit better with
tig's design.

Also, I currently re-run the `git diff` command every time the diff view is opened from blame, since I couldn't figure out an easier way to have the view load without using the line it was previously opened at.


src/diff.c
```diff
@@ -452,6 +453,14 @@ diff_save_line(struct view *view, struct diff_state *state, enum open_flags flag
 			state->pos = view->pos;
 		}
 	}
+	if (flags & OPEN_BLAMED_LINE) {
+		const char *file = filename_from_blame;
+		if (file) {
+			state->file = get_path(file);
+			state->lineno = lineno_from_blame;
+			state->pos.offset = 5; // TODO: Should be half screen
```
ilyagr, line 461: I haven't yet figured out how this works.

src/blame.c
```diff
@@ -451,7 +455,12 @@ blame_request(struct view *view, enum request request, struct line *line)
 			if (diff->pipe)
 				string_copy_rev(diff->ref, NULL_ID);
 		} else {
-			open_diff_view(view, flags);
+			if (filename_from_blame) {
+				free(filename_from_blame);
```
ilyagr, line 459: Somehow, I don't remember the pattern of "free the memory all but one time" in Kerninghan and Ritchie. 😢

Perhaps I should write to them and they'll put it into the next edition.

src/blame.c
```diff
@@ -451,7 +455,12 @@ blame_request(struct view *view, enum request request, struct line *line)
...
+			}
+			filename_from_blame = strdup(state->filename);
```
ilyagr, line 461: This does not work correctly if the file was renamed since the blamed line. TODO: fix (use filename from blame->commit), test.

src/blame.c
ilyagr: OK, I think I know how to fix this: simply access the global `blame_view` from `diff.c` to get this info. I may or may not get to it before I hear something back about whether the project is interested in this functionality.

include/tig/blame.h
```diff
@@ -17,6 +17,8 @@
 #include "tig/view.h"
 
 extern struct view blame_view;
+extern char *filename_from_blame;
+extern unsigned long lineno_from_blame;
```
ilyagr, line 21: I think creating these two globals isn't actually necessary. This information is also stored inside `blame_view`, though it's a bit fussy to extract it (see also a similar comment I made below).

koutcher:
Thanks, the experience is really good. As you expected, your implementation is not fully in line with Tig way of doing things, I pushed a commit with my review directly to your branch.

src/blame.c
```diff
@@ -455,12 +451,9 @@ blame_request(struct view *view, enum request request, struct line *line)
 			if (diff->pipe)
 				string_copy_rev(diff->ref, NULL_ID);
 		} else {
-			if (filename_from_blame) {
-				free(filename_from_blame);
-			}
-			filename_from_blame = strdup(blame->commit->filename);
-			lineno_from_blame = blame -> lineno;
-			open_diff_view(view, flags | OPEN_RELOAD | OPEN_BLAMED_LINE);
+			string_ncopy(view->env->file, blame->commit->filename, strlen(blame->commit->filename));
```
ilyagr, line 454: Is it impossible for strlen(blame->commit->filename) to be  more than SIZEOF_STR long?

ilyagr (commented):
Glad you like it! Thanks for all the fixes!

I think the test you removed was mostly testing for forgetting to set the `OPEN_RELOAD` flag (which I of course did in the first versions of the patch).

src/diff.c
```diff
@@ -543,6 +543,15 @@ diff_read(struct view *view, struct buffer *buf, bool force_stop)
 			}
 		}
 
+		if (view->env->blame_lineno) {
```
ilyagr, line 546: Just curious, why does the code go here? Is it just stylistic?

Also, `state->pos` is really bizarre :)

src/blame.c
koutcher: Maybe not, but the diff view already relies on it.

src/diff.c
koutcher: Agreed but it's almost as old as Tig. We need to run this code after the file has been read in order to know how many lines there are, so we have to be in the `if (!buf) {`.

Daghall:
Hmm, this is not improvement for me, personally. I usually just look at the commit message, rather than the diff, and the "old way" was perfect for my use case. 😊 

Is there some way to make it work as it did before?

I've tried
```
bind blame <Tab> enter; move-first-line
```

But that doesn't work (because the loading is asynchronous?), and I get the message `Only one view is displayed`.

koutcher:
`bind blame <Enter> view-diff`

Daghall:
> `bind blame <Enter> view-diff`

Not exactly what I wanted, but definitely good enough. Thank you! 😀

PR #1403, opened 2026-02-11
keybinding: adding C shortcut in diff view

spomata:
While in diff view, 'C' will copy the currently
highlighted commit SHA using xclip.


koutcher:
Thanks, the use of xclip is not portable so it can't be added to the default tigrc. Anyone needing it can add it easily to their own .tigrc, it is even show as an example in the man page.

PR #1413, opened 2026-03-30
support option append

liangkarl:
Adding append support for string-based options makes tigrc configuration more composable and less error-prone. Today, options such as main-options must be rewritten in full whenever a user wants to add one more argument, which leads to duplication and makes small customizations harder to maintain. With append support, users can keep a shared base configuration and extend it incrementally in later config blocks or local overrides. This improves readability, reduces accidental loss of existing arguments, and makes configuration layering much easier.


koutcher:
Good idea, thanks.

PR #1416, opened 2026-04-03
Fix blob view edit not writing changes to working directory

### LINKED ISSUE #1415: [BUG]Editing file via e in blob view does not modify the working directory file
Description

  When using the file finder (f) to locate and open a file in blob view, pressing e to edit
  the file in an external editor (nvim) does not result in changes to the working directory
  file.

  Steps to Reproduce

  1. Open tig
  2. Press f to find a file
  3. Press Enter to open the file in blob view
  4. Press e to edit in external editor (nvim)
  5. Make changes and save with :wq
  6. Return to tig, press s to enter status view
  7. Expected: the file appears under "Changes not staged for commit"
  8. Actual: no changes detected, status view shows no modifications

  Environment

  - OS: macOS (Darwin 25.4.0)
  - tig version: (run tig --version to get this)
  - Editor: nvim (set via $EDITOR)
  - Git version: (run git --version to get this)

  Workaround

  Editing from the status view (s) with e works correctly and modifies the working directory
  file as expected.

  Notes

  It appears that e in blob view may be opening a git object or temporary file rather than
  the actual working directory file. The status bar in nvim (configured with set
  statusline+=%F) could help verify which file path is being opened.

phpmac:
## Summary

When using the file finder (`f`) to open a file in blob view, pressing `e` to edit opens a **temporary file** instead of the working directory file. Changes are silently lost when the editor closes.

## Root Cause

In `blob_request()` (`src/blob.c:152-157`), when `state->file` is NULL (which is the case when entering blob view from the file finder), the code unconditionally calls `open_blob_editor()`:

```c
case REQ_EDIT:
    if (state->file)
        open_editor(state->file, ...);       // working dir file
    else
        open_blob_editor(view->vid, ...);    // temp file (deleted after edit!)
```

`open_blob_editor()` (`src/tree.c:336-360`) works by:
1. Running `git cat-file blob <id>` to write blob content to a temp file (`/tmp/tigblob.XXXXXX.xxx`)
2. Opening the editor on the temp file
3. Calling `unlink()` to **delete the temp file** immediately after the editor closes

Since the edit targets a temporary file that gets deleted, all changes are lost.

## Contrast with Tree View

The tree view (`src/tree.c:378-386`) handles this correctly by checking whether the current commit is HEAD:

```c
case REQ_EDIT:
    if (!is_head_commit(view->vid))
        open_blob_editor(entry->id, entry->name, 0);  // historical commit → temp file
    else
        open_editor(view->env->file, 0);               // HEAD → working dir file
```

## Fix

Add the same `is_head_commit()` check to blob view. When viewing a file at HEAD (which is always the case from the file finder), edit the working directory file directly:

```c
case REQ_EDIT:
    if (state->file)
        open_editor(state->file, ...);
    else if (is_head_commit(state->commit))
        open_editor(view->env->file, ...);              // ← new: HEAD → working dir
    else
        open_blob_editor(view->vid, ...);               // historical → temp file
```

## Steps to Reproduce

1. Open tig
2. Press `f` to open file finder
3. Select a file and press Enter to enter blob view
4. Press `e` to edit
5. Make changes and save (`:wq`)
6. Return to tig, press `s` for status view
7. **Expected**: file appears under "Changes not staged for commit"
8. **Actual**: no changes detected

Fixes #1415

### DIFF
--- a/src/blob.c
+++ b/src/blob.c
@@ -152,6 +152,8 @@ blob_request(struct view *view, enum request request, struct line *line)
 	case REQ_EDIT:
 		if (state->file)
 			open_editor(state->file, (line - view->line) + 1);
+		else if (is_head_commit(state->commit))
+			open_editor(view->env->file, (line - view->line) + 1);
 		else
 			open_blob_editor(view->vid, basename(view->ref), (line - view->line) + 1);
 		return REQ_NONE;


phpmac:
A quick note on the use case: this fix makes the `f` → Enter → `e` workflow actually useful for quickly finding and editing files without leaving tig. Previously you had to go through the status or tree view to edit files, which defeats the purpose of the file finder as a fast navigation tool.

Also, Ghostty + Claude Code + tig is an unstoppable combo.

koutcher:
Thanks. I wouldn't call it a bug since edit in the blob view was never meant to do this, but why not, with the extra condition that the file was not modified otherwise we lose the access to the original blob.

PR #1419, opened 2026-04-21
chocolate-theme: Fix cursor color on selecting commit in main view.

siddhpant:
Otherwise we have a bright yellow cursor on pressing enter in main view. White text on bright yellow cursor is very hard to read.

Without:

image

<br>

With:

image


koutcher:
Thanks. As a general rule, no PR aiming at changing the colors will be accepted, this one being the exception as it fixes an omission when `cursor-blur` was introduced.

To follow the same logic as the default theme, please make it `color cursor-blur		white	101` without `bold` to highlight the change of view.

siddhpant:
Done.

PR #1424, opened 2026-05-21
Fix slow tig status in large repo when given a subdirectory path

apawn:
## Summary

`git status `is quite slow in large repositories like chromium, but  `git status <path>` is much faster because it limits the scan to the given pathspec. while `tig status <path>` was as slow as `tig status`.

This change makes the status view pass path arguments to the underlying Git commands and skip the full index refresh when path filtering is active.

## Changes

- Pass `%(fileargs)` to `git diff-index`, `git diff-files`, and `git ls-files` when loading the status view
- Enable file argument expansion in `status_run()` (same mechanism as log/diff views)
- Skip `git update-index --refresh` when a path filter is active (this was the main bottleneck in large repos)
- Add `VIEW_FILE_FILTER` to the status view
- Add `test/status/file-filter-test`

## Result
- Before: 
   It used to take ages—around 37 seconds—looking like the app completely froze.
20260522161217_rec_

- After:  Reduced to ~1 second, drastically improving the user experience for this widely used feature.
20260522161148_rec_

## Test plan

- [x] `make test/status/file-filter-test`
- [x] In a large repo, verify `tig status subdir` is much faster than `tig status`
- [x] Verify `tig status subdir` shows the same changes as `git status subdir`
- [x] Verify `tig status` (without path) still works as before

### DIFF
--- a/include/tig/git.h
+++ b/include/tig/git.h
@@ -38,10 +38,10 @@
 /* Don't show staged unmerged entries. */
 #define GIT_DIFF_STAGED_FILES(output_arg) \
 	"git", "diff-index", (output_arg), "%(cmdlineargs)", "--diff-filter=ACDMRTXB", \
-		"-C", "--cached", "HEAD", "--", NULL
+		"-C", "--cached", "HEAD", "--", "%(fileargs)", NULL
 
 #define GIT_DIFF_UNSTAGED_FILES(output_arg) \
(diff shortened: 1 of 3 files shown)


koutcher:
Thanks. Merged in 38b53b68917244007a1cdcb83c87fe70f0191c63. When tig was started with `tig subdir` and you want the "full" status, press `%` to toggle the `file-filter` mode.

PR #1444, opened 2026-08-19
Update git information after running ':!'

liangkarl:
The current program won't update git information after running command with ':!'.

For example, a simple command running under the main view.
```:!git tag test```
The main view wasn't update until manual refresh.

Actually, any commands issued by ':!' would have the same problem, so this commit is to fix the problem.

### DIFF
--- a/include/tig/view.h
+++ b/include/tig/view.h
@@ -157,6 +157,7 @@ struct view {
 	time_t update_secs;
 	struct encoding *encoding;
 	bool unrefreshable;
+	bool watch_after_command;
 	struct watch watch;
 
 	/* Private data */
--- a/src/pager.c
+++ b/src/pager.c
@@ -21,6 +21,7 @@
 #include "tig/view.h"
 #include "tig/draw.h"
 #include "tig/diff.h"
+#include "tig/watch.h"
 
 /*
  * Pager backend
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by koutcher** [comment]
Why not simply:
```patch
--- a/src/prompt.c
+++ b/src/prompt.c
@@ -998,6 +998,7 @@ run_prompt_command(struct view *view, const char *argv[])

                        next->dir = NULL;
                        open_pager_view(view, OPEN_PREPARED | OPEN_WITH_STDERR);
+                       watch_update(WATCH_EVENT_AFTER_COMMAND);
                }

        } else if (!strcmp(cmd, "goto")) {
```

As a workaround, `:exec !git tag test` behaves as you expect

liangkarl:
Because `open_pager_view()` starts the command asynchronously and returns before the command has finished.
In this path, `open_pager_view()` opens Tig's pager, and `pager_open()` calls `begin_update()`.  `begin_update()` launches the process through `view_exec()` and then returns after setting up the incremental update machinery; it does not wait for the child process to exit. 

The name `WATCH_EVENT_AFTER_COMMAND` carries lifecycle semantics. The watch implementation treats it specially: it resets watch state, marks every registered trigger as changed, and records new timestamps. Doing that at process startup would violate those semantics and could record timestamps before the relevant filesystem changes occur, implying that it can refresh from an incomplete repository state.

About the use of `watch_after_command`
Not every use of the pager represents a :! command. The pager can also display existing input and other generated content. Consequently, pager EOF alone is insufficient to decide whether Tig should emit an after-command event.

koutcher:
Right, thanks. The pager view only gets arguments when running a prepared command, so we could get rid of the watch_after_command variable and have:
```
               if (view->argv)
                       watch_update(WATCH_EVENT_AFTER_COMMAND);
```

liangkarl:
Yes. That also works. Maybe this would be safer.
```
	if (view->argv && view->argv[0])
		watch_update(WATCH_EVENT_AFTER_COMMAND);
```

PR #1452, opened 2026-09-12
Preserve empty quoted arguments in bind commands

### LINKED ISSUE #1410: [Bug]: empty quoted parameter completely skipped in `.tigrc` `bind` command
Here's a `.tigrc` to reproduce:

```rc
bind main 1 >sh -c "./print_args_count --status '' ''"
bind main 2 >./print_args_count --status '  ' '    '
bind main 3 >./print_args_count --status '  ' '    '
bind main 4 >./print_args_count --status '' ''
bind main 5 >./print_args_count --status "" ""
```

`./print_args_count `:

```bash
#!/usr/bin/env bash
set -euf -o pipefail

echo "$#"
exit 1
```

Results:

 -  `1` ->  3
 -  `2` ->  3
 -  `3` ->  3
 -  `4` ->  1
 -  `5` ->  1

See there that the 4 and 5 cases will count 1 argument when I should have counted 3.

 Workaround is to always wrap commands in `sh -c ""`.
However, it would be really nice not to have to do that when not using shell features...

vulragrag-star:
Fixes #1410

Empty quoted parameters in `bind` commands were dropped when formatting
the run-request argv. For example:

```
bind main 1 >./print_args_count --status '' ''
```

ran the command with only `--status` (argc 1) instead of three arguments.

`format_append_arg` now keeps argv elements that were empty before
`%(var)` expansion, while still omitting empty results from expanding
`%(vars)`.

### Test plan
- [x] `make test/tigrc/empty-quoted-args-test` (new regression)
- [x] `make test/tigrc/quote-test`
- [x] `make test/tigrc/escape-var-test`


vulragrag-star:
AppVeyor was failing on status/untracked screen asserts. Root cause was in our patch: preserving every empty argv element also kept the "" placeholders used by internal git/status command templates (so `git ls-files` got spurious empty args and showed no untracked files).

Gated empty-arg preservation behind `argv_flag_preserve_empty` and only set that for run-request formatting. Internal templates still omit "" via `argv_appendn`.

Addressed in 7cb67a8e.

